PR #1, opened 2015-07-16
Initial ssh provider support

iNecas:


README.md
```diff
@@ -0,0 +1,45 @@
...
+To configure this plugin you can use template from settings.d/ssh.yml.example.
+You must place ssh.yml config file (based on this template) to your 
+smart-proxy config/settings.d/ directory.
+
+The simplest thing one can do is just to trigger a command:
+
+```
+curl http://my-proxy.example.com:9292/ssh/command \
+     -H 'Content-Type: application/json' \
+     -d '{"id":"123","effective_user":"root","script":"/usr/bin/ls","host":"my-host.example.com}'
```
jlsherrill: So playing with this, this command doesn't seem to work.  I can't seem to identify where that route would even be getting captured.  I'm getting a 404.  Is there a commit missing?  

Also, you're missing -X POST from that command.

jlsherrill:
Would like to see rubocop here initially as well :)

README.md
iNecas: Sorry for that, I've forgot to update the README after I've connected the tasks with the server, updated now.

iNecas:
Rubocop added

jlsherrill:
One thing i noticed, on a system where the specified ssh key was not authorized (and it fell back to password auth), the terminal running my proxy got prompted for:

root@localhost's password:

and i could actually enter passwords.  I'm guessing std input needs to be redirected somehow?  It seemed like it was just going to sit there forever (with the task stuck in pending).

jlsherrill:
One option would be to disable password auth within net-ssh (which looks possible)

jlsherrill:
Doesn't look like the scripts get cleaned up out of /tmp/foreman-proxy-ssh/server/ or /tmp/foreman-proxy-ssh/client/

Do we want to do that?

jlsherrill:
When trying to ssh to a machine that has ssh turned off, it seems to retry forever.

smart_proxy_remote_execution_ssh.gemspec
```diff
@@ -0,0 +1,36 @@
...
+    Ssh remote execution provider for Foreman Smart-Proxy
+  EOS
+
+  gem.files         = Dir['{bundler.d,lib,settings.d}/**/*', 'LICENSE', 'Gemfile']
+  gem.test_files    = gem.files.grep(%r{^(test|spec|features)/})
+  gem.require_paths = ["lib"]
+  gem.license = 'GPLv3'
+
+  gem.add_development_dependency "bundler", "~> 1.7"
+  gem.add_development_dependency "rake", "~> 10.0"
```
stbenjam: I end up getting this because the smart proxy already requires rake:

```
The gemspec at
/home/stbenjam/git/smart_proxy_remote_execution_ssh/smart_proxy_remote_execution_ssh.gemspec is not
valid. The validation error was 'duplicate dependency on rake (~> 10, development), (~> 10.0) use:
    add_runtime_dependency 'rake', '~> 10', '~> 10.0'
```

jlsherrill: I think its actually that *this* gemspec requires it twice  line 28 and 23

jlsherrill:
I'm fine fixing the issues i mentioned in later Pull requests, if you want to get this one merged.  Everything else looks good to me :+1:

smart_proxy_remote_execution_ssh.gemspec
stbenjam: Oh good catch! It's the same over in smart_proxy_dynflow too

iNecas:
@jlsherrill thanks for testing the things, I want to work more on the keys settings (both identity keys, as well as server keys known hosts today), so we can merge after this changes are in

iNecas:
I've added the support for simple key management (see the commit message) and fixed the rake. I will open the subsequent tasks as issues and address them separately, after merging this one

iNecas:
@jlsherrill I've fixed the issue with the password being asked when the public key auth fails

PR #2, opened 2015-08-14
Postpone the initialization

iNecas:
We can't initialize the plugin sooner than the smart proxy forks in 
production, because otherwise we loose the file descriptors we need to open at 
the initialization phase.

Therefore, we use hooking mechanism that smart_proxy_dynflow provides


iNecas:
requires https://github.com/theforeman/smart_proxy_dynflow/pull/4

stbenjam:
ACK

PR #3, opened 2015-08-19
Fixes needed for packaging

stbenjam:


stbenjam:
Does it make sense to include the Gemfile in the gem itself? It was only intended for development use, and I really didn't want that to have to end up in the RPM package.

jlsherrill:
@stbenjam i think it was likely not intentional.  ACK

stbenjam:
Needed another change, ssh.yml doesn't follow convention.

jlsherrill:
ACK

PR #5, opened 2015-08-26
Fixes #11553 - introduce session actors to avoid blocking on opening connections

iNecas:
The commands now work more independently, not having wait while other 
connections are getting opened.

### DIFF
--- a/lib/smart_proxy_remote_execution_ssh/command_update.rb
+++ b/lib/smart_proxy_remote_execution_ssh/command_update.rb
@@ -1,0 +1,73 @@
+module Proxy::RemoteExecution::Ssh
+  # update sent back to the suspended action
+  class CommandUpdate
+    attr_reader :buffer, :exit_status
+
+    def initialize(buffer)
+      @buffer = buffer
+      extract_exit_status
+    end
+
+    def extract_exit_status
+      @buffer.delete_if do |data|
+        if data.is_a? StatusData
+          @exit_status = data.data
+          true
+        end
+      end
+    end
+
+    def buffer_to_hash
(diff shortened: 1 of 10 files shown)


lib/smart_proxy_remote_execution_ssh/command_update.rb
```diff
@@ -0,0 +1,73 @@
+module Proxy::RemoteExecution::Ssh
+  # update sent back to the suspended action
+  class CommandUpdate
+    attr_reader :buffer, :exit_status
+
+    def initialize(buffer)
+      @buffer      = buffer
```
ares: nit: extra whitespace?

lib/smart_proxy_remote_execution_ssh/session.rb
```diff
@@ -0,0 +1,169 @@
...
+      self.parent
+    end
+
+    def start_termination(*args)
+      super
+      close
+      finish_termination
+    end
+
+    protected
```
ares: I know it's copied, but shouldn't this be private?

lib/smart_proxy_remote_execution_ssh/command_update.rb
iNecas: White-space commit pushed (I've also updated some alignations on =>?, as I know you don't like it:)

lib/smart_proxy_remote_execution_ssh/session.rb
iNecas: right, updated

ares:
ACK

iNecas:
Squashed, merging

PR #6, opened 2015-08-31
Fixes #11591 - propagate script exit codes

iNecas:


lib/smart_proxy_remote_execution_ssh/session.rb
```diff
@@ -30,7 +30,7 @@ def initialize_command
       end
       output_path = File.join(File.dirname(remote_script), 'output')
 
-      @connector.async_run("#{su_prefix}#{remote_script} | /usr/bin/tee #{output_path}") do |data|
+      @connector.async_run("set -o pipefail; #{su_prefix}#{remote_script} | /usr/bin/tee #{output_path}") do |data|
```
adamruzicka: This assumes the effective user's shell supports `pipefail`. Guess for now we should just mention it somewhere in the docs so the users won't be surprised by exit codes being wrong and/or messages like `set: Illegal option -o pipefail` in the output in case they're using something like `dash`.

ohadlevy: if its base you can also uses PIPESTATUS, probably can test for the user shell or if it supports pipefail.

iNecas: I've updated the PR with the most generic POSIX styl I've found, heavily inspired by http://unix.stackexchange.com/a/16709, tested on bash, dash, zsh.

lib/smart_proxy_remote_execution_ssh/session.rb
```diff
@@ -30,7 +30,14 @@ def initialize_command
       end
       output_path = File.join(File.dirname(remote_script), 'output')
 
-      @connector.async_run("#{su_prefix}#{remote_script} | /usr/bin/tee #{output_path}") do |data|
+      # pipe the output to tee while capturing the exit code
+      script = <<-SCRIPT
+      exec 4>&1
```
ares: when I use heredoc, I either remove indentation or indent it, this way I think it keeps 6 spaces in begining of each line so either 0 or 8?

ares:
I can't say I fully understand the patch but it works and fixes the issue :-)

lib/smart_proxy_remote_execution_ssh/session.rb
iNecas: Damn arescop :) fixed

ares:
:+1:

lib/smart_proxy_remote_execution_ssh/session.rb
ares: Sorry :-)

iNecas:
We have some issues with CI still, but tested locally and worked: the master should get green soon

PR #7, opened 2015-08-31
Fixes #11593 - handle encoding of the command output properly

iNecas:
Or at lest, enforcing the UTF-8 doesn't cause encoding issues, when the
command produces UTF characters.


ohadlevy:
tests?

iNecas:
Added

ares:
:+1: tested, fixes the issue

PR #10, opened 2015-10-05
refs #11634 - use ssh_user from task input

stbenjam:


lib/smart_proxy_remote_execution_ssh/command_action.rb
```diff
@@ -42,7 +42,7 @@ def rescue_strategy_for_self
     def command
       @command ||= Dispatcher::Command.new(:id               => input[:task_id],
                                            :host             => input[:hostname],
-                                           :ssh_user         => 'root',
+                                           :ssh_user         => input[:ssh_user] || 'root',
```
iNecas: While touching that, making the 'root' as config option would be nice as well

stbenjam: I'd actually be more inclined to remove the || case, and make foreman always send the relevant user name. Any reason to configure it on the proxy side?

iNecas: It felt close to the private keys setting, which we want to be able to keep on per-proxy. I'm ok with letting override both remote user and private key from the server, but would still like to make the possibility to keep defaults for the private keys (and remote user that should be used) here.

stbenjam: I hadn't pictured a case where the Foreman doesn't send a user name, though.

The biggest issue for me is cloud instances tend to use a non-privileged user particular to the image (cloud-user, fedora, vagrant, etc) whereas bare metal may use root or a non-privileged user (maybe one dedicated for Foreman's use).  So I considered it a host/host group property, not a property of the SSH key or the proxy.

That's the use case I was trying to solve for iteration 3.

Are we still considering some alternate cases where the ssh user is the current foreman user's? Or doing more key management?

I was hoping we'd stay simple with ssh keys and offer FreeIPA as an option for more granular options (assuming delegation does what we want).

Anyway, can we have a chat on Monday about what the requirements are, and what's possible for iteration 3?

iNecas: What we are currently looking is a way how to determine the effective user (via sudo) at the server side, but it could be similar to the remote user. Let's chat on monday about that

stbenjam: Huh? That's the effective user, it's not #11634, which is the ssh user.

iNecas:
ACK, tested and works like a charm!

PR #11, opened 2015-10-08
fixes #12108 - Require smart_proxy_dynflow 0.0.4

stbenjam:


iNecas:
Is there any code change in smart_proxy_dynflow 0.0.4, that the latest versions of smart_proxy_remote_execution_ssh requires? It seems the dependency should be between foreman plugins and smart proxy plugins, but that's not something we support to define right now

stbenjam:
Oh, I misunderstood the user's e-mail, nevermind then.

PR #12, opened 2015-10-26
fixes #12313 - remove unused item from settings yaml

stbenjam:
Also removed from installer https://github.com/theforeman/puppet-foreman_proxy/pull/201


iNecas:
ACK

PR #14, opened 2015-11-11
fixes #11955, #12422 - use /var/tmp and a unique path on clients

stbenjam:
Reason for /var/tmp was they actually did want them persisted on reboots, and also many places mount /tmp as noexec.

https://groups.google.com/forum/#!msg/foreman-users/yXFMiB324G0/6bTIMSaVBwAJ

### DIFF
--- a/lib/smart_proxy_remote_execution_ssh/session.rb
+++ b/lib/smart_proxy_remote_execution_ssh/session.rb
@@ -9,8 +9,8 @@ module Proxy::RemoteExecution::Ssh
       @clock = options[:clock] || Dynflow::Clock.spawn('proxy-dispatcher-clock')
       @logger = options[:logger] || Logger.new($stderr)
       @connector_class = options[:connector_class] || Connector
-      @local_working_dir = options[:local_working_dir] || '/tmp/foreman-proxy-ssh/server'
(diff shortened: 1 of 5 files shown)


lib/smart_proxy_remote_execution_ssh/session.rb
```diff
@@ -121,7 +121,7 @@ def local_command_file(filename)
     end
 
     def remote_command_dir
-      File.join(@remote_working_dir, @command.id)
+      "#{@remote_working_dir}-#{@command.id}"
```
iNecas: I wonder, if the `@remote_working_dir` shouldn't be just `/var/tmp`, as there wouldn not be `/var/tmp/foreman-proxy-ssh` directory created, but `/var/tmp/foreman-proxy-ssh-#{uuid}`. So it would be:

```ruby
File.join(@remote_working_dir, "foreman-proxy-ssh-command-#{@command.id}")
```

stbenjam: Thanks, that makes sense, I'll do that.  That's a bit long, how about `foreman-ssh-cmd-#{@command.id}`?

iNecas:
Tested and works great. ACK

PR #15, opened 2015-11-12
Fixes #12456 - handle the encoding in general, not just in stdout/stderr

iNecas:
We can run into similar issues with exceptions as well. Handling the encoding
on more high level.


stbenjam:
APJ

iNecas:
the jenkins is failling just because it still had previous version of the PR on ruby 2.1

iNecas:
[test]

PR #16, opened 2015-11-30
Refs #12489 - enable sudo

iNecas:


iNecas:
The proxy now allows to use sudo to switch to the effective user from the remote one

stbenjam:
Looks good, haven't tried it yet.

How will the Foreman side decide to use sudo or not?

iNecas:
Similar to the mechanism we have son the ssh_user - using a host param, default to settings

iNecas:
The PR on foreman_remote_execution is finally here https://github.com/theforeman/foreman_remote_execution/issues/80

lib/smart_proxy_remote_execution_ssh/dispatcher.rb
```diff
@@ -7,18 +7,19 @@ module Proxy::RemoteExecution::Ssh
...
-        @ssh_user         = data[:ssh_user]
-        @effective_user   = data[:effective_user]
-        @script           = data[:script]
-        @host_public_key  = data[:host_public_key]
-        @suspended_action = data[:suspended_action]
+        @id                    = data[:id]
+        @host                  = data[:host]
+        @ssh_user              = data[:ssh_user]
+        @effective_user        = data[:effective_user]
+        @effective_user_method = data[:effective_user_method] || 'su'
```
stbenjam: Can we make the default 'sudo' to match Foreman's default? Or was there a reason for picking su?

stbenjam: Nevermind, I see, we used to use `su` so this keeps the behavior of the proxy the same with older Foremans.

stbenjam:
Works well, ACK from me

iNecas:
Thanks @stbenjam for review

PR #17, opened 2015-12-21
Fixes #12902 - Handle net-ssh dependency differently

stbenjam:
Some OS platforms will move on to packaging net-ssh 3.x only


ares:
ACK

PR #18, opened 2016-01-04
Fixes #12786 - Load sinatra earlier to ensure it's available

dLobatog:
Before this patch, Sinatra could be referenced before it's loaded.
This patch simply allows it to load when the plugin is loaded.


stbenjam:
Thanks @dLobatog!

stbenjam:
ACK

PR #19, opened 2016-01-19
fixes #13274 - the smart proxy service can now load

ohadlevy:


stbenjam:
Do you know why this fixes it? I can't reproduce locally on any ruby version I try. I can see why we'd need smart_proxy_dynflow, but why the version?

ohadlevy:
@stbenjam what loads the version? since its not required anywhere, there is no auto loader enabled.

stbenjam:
It was moved to the after_activation hook.  https://github.com/theforeman/smart_proxy_remote_execution_ssh/blob/master/lib/smart_proxy_remote_execution_ssh/plugin.rb#L15

Is it needed before? I think we should probably only require it in one place

ohadlevy:
not sure how it works TBH but since the plugin class calls version, it cant
be required afterwards. I assume the same for dynflow?

stbenjam:
Just looking at other examples (like [this one](https://github.com/theforeman/smart_proxy_dns_powerdns/blob/master/lib/smart_proxy_dns_powerdns/dns_powerdns_plugin.rb)), plugin requires version itself.

Then smart_proxy_dynflow require could go where you put it, and we remove the 2 from the after_activation hook?

I don't really have any strong preference, just want to make sure it's in a logical place and fixes the ordering issues.

ohadlevy:
TBH I dont mind at all, this simply fixed the issue for me.

iNecas:
I haven't found better place than in this PR. Reproduced, tested, worked, merged

stbenjam:
Cool, thanks @iNecas !

PR #20, opened 2016-01-31
Fixes #13491 - handle ssh disconnection

iNecas:
For example, when rebooting as part for the script.

### DIFF
--- a/lib/smart_proxy_remote_execution_ssh/session.rb
+++ b/lib/smart_proxy_remote_execution_ssh/session.rb
@@ -55,6 +55,10 @@ module Proxy::RemoteExecution::Ssh
           finish_command
         end
       end
+    rescue Net::SSH::Disconnect => e
+      @command_buffer.concat(CommandUpdate.encode_exception("Failed to refresh the connector", e, true))
+      refresh_command_buffer
+      finish_command
     rescue => e
       @command_buffer.concat(CommandUpdate.encode_exception("Failed to refresh the connector", e, false))
     ensure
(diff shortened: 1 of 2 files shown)


iNecas:
I've added a simple detection of expected disconnections based on stdout. The plan is to seed `power action` template that will produce such output so that it just works (and the user doesn't have to know exactly what's the expected message to make the tasks green)

stbenjam:
> The plan is to seed power action template that will produce such output so that it just works (and the user doesn't have to know exactly what's the expected message to make the tasks green)

Would it be possible to integrate into Foreman's power controls somehow? Not so familiar with this part but there is some BMC integration, and maybe there could be an alternative REX/SSH provider using the developer API.

iNecas:
For now, I've added a job template for the power management control: https://github.com/iNecas/foreman_remote_execution/commit/9ec393c1c04a744fa0627ed5731a5bce4925f0a5. Yes, the developer API could be used to intergrate it into the BMC, but I would prefer having a seperate issue opened for this.

lib/smart_proxy_remote_execution_ssh/session.rb
```diff
@@ -65,6 +76,7 @@ def refresh
     def refresh_command_buffer
       @logger.debug("command #{@command} got new output: #{@command_buffer.inspect}")
       command_update = CommandUpdate.new(@command_buffer)
+      check_expecting_disconnect
```
adamruzicka: What's the point of `check_expecting_disconnect` here?

iNecas: We need to determine, if the disconnection was expected (such as when calling `reboot`) or it was an error. We need to do it on two places, because the `@command_buffer` could be filled in this, or previous iteration.

adamruzicka: Oh, I see, thanks for the explanation

adamruzicka:
Works as expected, ACK

iNecas:
Issue for the BMC provider filled here http://projects.theforeman.org/issues/13688

PR #21, opened 2016-02-02
Refs #13523 - fix requiring code in tests

iNecas:


iNecas:
We need https://github.com/theforeman/smart_proxy_dynflow/pull/10 first for tests to even run again.

test/test_helper.rb
```diff
@@ -13,6 +13,11 @@
 # DYNFLOW_TESTING_LOG_LEVEL = 0 # for debugging
 require 'smart_proxy_dynflow/testing'
 require 'smart_proxy_remote_execution_ssh'
+require 'smart_proxy_remote_execution_ssh/connector'
+require 'smart_proxy_remote_execution_ssh/command_update'
+require 'smart_proxy_remote_execution_ssh/dispatcher'
+require 'smart_proxy_remote_execution_ssh/command_action'
+require 'smart_proxy_remote_execution_ssh/api'
```
stbenjam: Could we just use globbing to include everything? Next time we add something we might forget to put it here.

iNecas: I preffer explicit requires over globbing, it makes it easier to find where some file/module/class gets loaded. Another problem might be with the ordering.

If we need it in the tests, spotting something new is missing should be easy (the tests should fail, otherwise it's not needed:)

adamruzicka:
Looks good to me

PR #22, opened 2016-04-21
API

adamruzicka:
API part of taking dynflow out of the smart proxy to separate process, which lives [here](https://github.com/adamruzicka/smart_proxy_remote_execution_ssh/tree/core) until I figure out a proper name for it


smart_proxy_remote_execution_ssh.gemspec
```diff
@@ -14,7 +14,8 @@ Gem::Specification.new do |gem|
     Ssh remote execution provider for Foreman Smart-Proxy
   EOS
 
-  gem.files            = Dir['{bundler.d,lib,settings.d}/**/*', 'LICENSE', 'README.md']
+  gem.files            = Dir['lib/smart_proxy_remote_execution_ssh.rb', 'LICENSE', 'README.md',
+                             '{bundler.d,lib,settings.d}/**/*']
```
ares: this should not contain the whole lib but `lib/smart_proxy_remote_execution_ssh/**/*` and `lib/smart_proxy_remote_execution_ssh.rb` only, otherwise smart_proxy_remote_execution_ssh plugin tries to load the core as well

smart_proxy_remote_execution_ssh_core.gemspec
```diff
@@ -0,0 +1,36 @@
...
+  gem.add_development_dependency "rake", "~> 10.0"
+  gem.add_development_dependency('minitest')
+  gem.add_development_dependency('mocha', '~> 1')
+  gem.add_development_dependency('webmock', '~> 1')
+  gem.add_development_dependency('rack-test', '~> 0')
+  gem.add_development_dependency('rubocop', '0.32.1')
+
+  gem.add_runtime_dependency('smart_proxy_dynflow_core', '>= 0.0.7')
+
+  gem.add_runtime_dependency('net-ssh')
```
ares: we should pin this to '< 3', since 3+ is not 1.9 comptabile

smart_proxy_remote_execution_ssh_core.gemspec
```diff
@@ -0,0 +1,36 @@
...
+  gem.version       = Proxy::RemoteExecution::Ssh::Core::VERSION
+  gem.authors       = ['Ivan Nečas']
+  gem.email         = ['inecas@redhat.com']
+  gem.homepage      = "https://github.com/theforeman/smart_proxy_remote_execution_ssh"
+  gem.summary       = 'Ssh remote execution provider for Foreman Smart-Proxy'
+  gem.description   = <<-EOS
+    Ssh remote execution provider for Foreman Smart-Proxy
+  EOS
+
+  gem.files            = Dir['lib/smart_proxy_remote_execution_ssh_core', 'LICENSE', 'README.md',
```
ares: missing `.rb`

lib/smart_proxy_remote_execution_ssh.rb
```diff
@@ -5,7 +5,7 @@
 module Proxy::RemoteExecution
```
ares: this needs to be broken to
```
module Proxy
  module RemoteExecution
```
because if we run it standalone, Proxy does not exist

iNecas:
I've squashed the commits, pushed to master and release 0.1.0 version for both proxy and core gems

PR #23, opened 2016-06-02
Fixes #14901 - Fix OpenStruct issues

adamruzicka:


iNecas:
Lets see how that works http://ci.theforeman.org/job/test_proxy_plugin_pull_request/113/

iNecas:
Still some issues to get fixed http://ci.theforeman.org/job/test_proxy_plugin_pull_request/113/ruby=2.1/console

iNecas:
@adamruzicka could you look at the test failures (including those that cause the master to fail in jenkins right now http://ci.theforeman.org/job/test_proxy_plugin_matrix/291/ruby=2.1/console) to get this in?

adamruzicka:
We're [green](http://ci.theforeman.org/job/test_proxy_plugin_pull_request/124/)

Gemfile
```diff
@@ -8,6 +8,13 @@ group :development do
   gem 'pry'
 end
 
+if RUBY_VERSION < "2.0"
+  gem 'json', '< 2.0.0'
```
dLobatog: json should be `~> 1.8.2` IMO, 1.7.x had one of the worst Ruby CVEs https://www.ruby-lang.org/en/news/2013/02/22/json-dos-cve-2013-0269/

Gemfile
```diff
@@ -8,6 +8,13 @@ group :development do
...
+  gem 'net-ssh', '<= 2.10'
```
dLobatog: '~> 2.9.0' pins it better I think, there are a few things incompatible & warnings with 2.5.1

Gemfile
```diff
@@ -8,6 +8,13 @@ group :development do
...
+  gem 'rest-client', '< 1.7'
```
dLobatog: This has nasty implications...http://projects.theforeman.org/issues/13698 can we get 1.8.x at least? https://github.com/rest-client/rest-client/blob/1.8.x/rest-client.gemspec

dLobatog:
:+1: Thanks for the updates, they're much needed. Can you take a look at the Gemfile comments? I think the only one that might require some testing is the rest-client one

adamruzicka:
@dLobatog Thanks for looking into this. I'm not sure how `rubygems` pick the correct version, what made sense to me was for it to 'pick the latest version within the constraint', but I guess it won't do any harm if I switch them to be stricter.

Also the changes to the version pins would be needed in [smart_proxy_dynflow](https://github.com/theforeman/smart_proxy_dynflow) too, the Gemfile there handles it the same way

Gemfile
iNecas: I don't think it should be this library responisbility to make people use the latest working version (maybe somebody has some other reaosons to use 1.7.x). The reason of this contaraits are really to note which versions are known not working on a given RUBY_VERSION. Also, what if 1.9.x version gets released (here it's just a theory, but still), I think '< 2.0.0' serves better for this purposes here.

Gemfile
iNecas: Same as https://github.com/theforeman/smart_proxy_remote_execution_ssh/pull/23#discussion_r71832873

Gemfile
iNecas: I `'rest-client' < '2'` works, let's put that there instead

adamruzicka: done

Gemfile
dLobatog: It can be `~> 1.8` too, which would work with 1.9 -  I don't see why leaving any room for users to be the target of a serious CVE. If someone has a legitimate reason to still use 1.7 after that, let them open a bug and we'll handle it

Gemfile
dLobatog: `< 2` doesn't ensure anything, it allows ancient versions to be used. `~> 1.8` prevents at least a few issues and we know it's compatible with it

iNecas:
I rerun the [tests](http://ci.theforeman.org/job/test_proxy_plugin_pull_request/125/) and still :green_apple:

iNecas:
Merging, thanks @adamruzicka

PR #25, opened 2016-06-13
fixes #15385 - match foreman's version requirement on net-ssh

stbenjam:
As we're running in the same SCL, we need to be compat with the version
they're shipping.


ares:
untested but makes sense to me

stbenjam:
```
16:41 #theforeman-dev: | mhulan > stbenjam: consider my comment ACK :-)
```

Merging, thanks

PR #29, opened 2016-08-26
fixes #16321 - Invocation of script on target host

Klaas-:
I changed the acls for the script file to not be be writeable by others,
not sure if you can actually use to manipulate the job but better safe
than sorry

also I changed the invocation of the script on the receiver because
if you have noexec set to /tmp and /var/tmp bind mounted to tmp
like redhat recommends then execution of the script does not work.


Klaas-:
do I need to create two issues for this? :)

Klaas-:
closed as won't fix

PR #30, opened 2016-08-26
fixes #16322 - Insecure permissions on script file

Klaas-:
I changed the acls for the script file to not be be writeable by others,
not sure if you can actually use to manipulate the job but better safe
than sorry

Needs http://projects.theforeman.org/issues/16322 // #29 or different permissions (0555)

### DIFF
--- a/lib/smart_proxy_remote_execution_ssh_core/session.rb
+++ b/lib/smart_proxy_remote_execution_ssh_core/session.rb
@@ -162,7 +162,7 @@ module Proxy
 
         def cp_script_to_remote
           local_script_file = write_command_file_locally('script', sanitize_script(@command.script))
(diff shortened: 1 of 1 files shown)


iNecas:
After a discussion with others, it seems having the 0555 option is better, as it allows people to specify the interpreter of the job execution with a shebang and there doesn't seem to be that much higher security advantage with 0444, as one could just create the exececutable file inside the job and run that. If the issue is with setting executable flag on a tmp file, it should be probably handled by changing of directory location

iNecas:
We are extracting the code from the smart proxy to be usable by foreman core or foreman proxy. I've added a commit into https://github.com/theforeman/foreman_remote_execution/pull/203, that uses the 0555 permissions for the script there.

PR #32, opened 2017-05-29
Fixes #17514 - Allow enabling async ssh

adamruzicka:


settings.d/remote_execution_ssh.yml.example
```diff
@@ -4,3 +4,5 @@
 :local_working_dir: '/var/tmp'
 :remote_working_dir: '/var/tmp'
 # :kerberos_auth: false
+# :async_ssh: false
+# :runner_refresh_interval:
```
iNecas: Putting a comment here about the default could be nice for the end-user to know what to expect

iNecas:
Thanks @adamruzicka !

PR #33, opened 2017-06-21
Fixes #11936 - Add settings for kerberos auth

adamruzicka:


settings.d/remote_execution_ssh.yml.example
```diff
@@ -3,3 +3,4 @@
 :ssh_identity_key_file: '~/.ssh/id_rsa_foreman_proxy'
 :local_working_dir: '/var/tmp'
 :remote_working_dir: '/var/tmp'
+:kerberos_auth: false
```
iNecas: A hint: if we keep the setting commented, we have more control over the default setting in the code, if we wanted to enable it by default in future for some reason.

adamruzicka: Good point

iNecas:
Thanks @adamruzicka

PR #36, opened 2018-01-07
pin gem versions for older Rubies

mmoll:
also removes pinning for Ruby <2.0

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,15 +7,27 @@ group :development do
   gem 'pry'
 end
 
-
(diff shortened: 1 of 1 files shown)


Gemfile
```diff
@@ -7,15 +7,27 @@ group :development do
   gem 'pry'
 end
 
+group :test do
+if RUBY_VERSION < '2.1'
```
iNecas: Please indent this so that it's visible better it's in a group

iNecas:
Thanks @mmoll

PR #39, opened 2018-05-19
Do not use mocha methods outside of tests

ekohl:
This fixes the error in CI:

Mocha methods cannot be used outside the context of a test


iNecas:
Tested and works well.

PR #40, opened 2019-02-21
Add /ssh/session endpoint

mvollmer:


lib/smart_proxy_remote_execution_ssh/api.rb
```diff
@@ -4,6 +4,98 @@ class Api < ::Sinatra::Base
       get "/pubkey" do
         File.read(Ssh.public_key_file)
       end
+      post "/session" do
+        if env["HTTP_CONNECTION"] != "upgrade" or env["HTTP_UPGRADE"] != "raw" then
```
mvollmer: The value of the "upgrade" header is a list, and we need to check whether "raw" is in it.

lib/smart_proxy_remote_execution_ssh/api.rb
```diff
@@ -4,6 +4,98 @@ class Api < ::Sinatra::Base
...
+        methods.unshift('password') if params["ssh_password"]
+
+        ssh_options = { }
+        ssh_options[:port] = params["ssh_port"] if params["ssh_port"]
+        ssh_options[:keys] = [ key_file ] if key_file
+        ssh_options[:password] = params["ssh_password"] if params["ssh_password"]
+        ssh_options[:passphrase] = params[:ssh_key_passphrase] if params[:ssh_key_passphrase]
+        ssh_options[:keys_only] = true
+        ssh_options[:auth_methods] = methods
+        ssh_options[:verify_host_key] = :accept_new_or_local_tunnel
```
mvollmer: We should also accept a host key as a parameter and setup a suitable known_hosts file with it.

lib/smart_proxy_remote_execution_ssh/api.rb
```diff
@@ -4,6 +4,98 @@ class Api < ::Sinatra::Base
...
+      end
+
+      def ssh_on_socket(socket, command, ssh_user, host, ssh_options)
+        started = false
+        err_buf = ""
+        begin
+          Net::SSH.start(host, ssh_user, ssh_options) do |ssh|
+            channel = ssh.open_channel do |ch|
+              ch.exec(command) do |ch, success|
+                raise "could not execute command" unless success
```
mvollmer: We might need to catch this exception, or in general, make sure that the right thing happens when this code raises one.

lib/smart_proxy_remote_execution_ssh/api.rb
```diff
@@ -4,6 +4,98 @@ class Api < ::Sinatra::Base
...
+            end
+
+            channel.wait
+            if !started
+              socket.enqueue("Status: 400\r\n")
+              socket.enqueue("Connection: close\r\n")
+              socket.enqueue("\r\n")
+              socket.enqueue(err_buf)
+              socket.wait_for_pending_sends
+              socket.close
```
mvollmer: I am not really sure whether the socket is properly closed in all cases.

lib/smart_proxy_remote_execution_ssh/api.rb
adamruzicka: `then` is not really needed when the `if` spans multiple lines

mvollmer:
Will the connection from the Foreman master to the SmartProxy ever have to traverse proxies and/or caches?

iNecas:
Does this require puma?

mvollmer:
> Does this require puma?

Yes.  Without Puma, calling /ssh/session should fail cleanly since the socket hijacking will fail.  (But I haven't tested that yet.)

lib/smart_proxy_remote_execution_ssh/api.rb
mvollmer: Removed the `then`s, thanks.  (The first comment about the upgrade header is still open.)

lib/smart_proxy_remote_execution_ssh/api.rb
```diff
@@ -4,6 +4,98 @@ class Api < ::Sinatra::Base
...
+        if env['rack.hijack?']
+          env['rack.hijack'].call
+          socket = env['rack.hijack_io']
+        end
+        if !socket
+          return [ 501, "Internal error: request hijacking not available" ]
+        end
+
+        params = MultiJson.load(env["rack.input"].read)
+        key_file = ForemanRemoteExecutionCore.settings.fetch(:ssh_identity_key_file)
```
mvollmer: In a production setup, `ForemanRemoteExecutionCore` is not available in the proxy, I think.

iNecas: I guess we can use `Proxy::RemoteExecution::Ssh.private_key_file` from https://github.com/theforeman/smart_proxy_remote_execution_ssh/blob/master/lib/smart_proxy_remote_execution_ssh.rb#L24

lib/smart_proxy_remote_execution_ssh/api.rb
mvollmer: Done.

lib/smart_proxy_remote_execution_ssh/api.rb
mvollmer: It's now closed in all 'normal' cases.  It stays open when the code raises an exception after hijacking and before setting up the rescue block in ssh_on_socket.  I guess that is enough...

lib/smart_proxy_remote_execution_ssh/api.rb
mvollmer: Thanks! Done.

lib/smart_proxy_remote_execution_ssh/api.rb
```diff
@@ -4,6 +4,120 @@ class Api < ::Sinatra::Base
...
+        if env["HTTP_CONNECTION"] != "upgrade" or env["HTTP_UPGRADE"] != "raw"
+          return [ 401, "Invalid request: /ssh/session requires connection upgrade to 'raw'" ]
+        end
+        socket = nil
+        if env['rack.hijack?']
+          env['rack.hijack'].call
+          socket = env['rack.hijack_io']
+        end
+        if !socket
+          return [ 501, "Internal error: request hijacking not available" ]
```
adamruzicka: `unless $CONDITION` is preferred over `if !$CONDITION`, This whole block could be rewritten as

```ruby
return [501, "Internal error: request hijacking not available"] unless socket
```

On a side note, spaces just after the opening bracket (and just before the closing one) are not really all that common in ruby world

lib/smart_proxy_remote_execution_ssh/api.rb
```diff
@@ -4,6 +4,120 @@ class Api < ::Sinatra::Base
...
+        101
+      end
+
+      def ssh_on_socket(socket, command, ssh_user, host, ssh_options)
+        started = false
+        err_buf = ""
+        socket.extend(Net::SSH::BufferedIo)
+
+        send_start = -> {
+          if !started
```
adamruzicka: `unless started`

lib/smart_proxy_remote_execution_ssh/api.rb
```diff
@@ -4,6 +4,120 @@ class Api < ::Sinatra::Base
...
+          Net::SSH.start(host, ssh_user, ssh_options) do |ssh|
+            channel = ssh.open_channel do |ch|
+              ch.exec(command) do |ch, success|
+                raise "could not execute command" unless success
+
+                ssh.listen_to(socket)
+
+                ch.on_process do
+                  if socket.available > 0
+                    ch.send_data(socket.read_available)
```
adamruzicka: This could be rewritten as
```ruby
ch.send_data(socket.read_available) if socket.available.positive?
```

lib/smart_proxy_remote_execution_ssh/api.rb
```diff
@@ -4,6 +4,120 @@ class Api < ::Sinatra::Base
...
+                  end
+                  if socket.closed?
+                    ch.close
```
adamruzicka: This could be a oneliner too

lib/smart_proxy_remote_execution_ssh/api.rb
```diff
@@ -4,6 +4,120 @@ class Api < ::Sinatra::Base
...
+
+                ch.on_data do |ch2, data|
+                  send_start.call
+                  socket.enqueue(data)
+                end
+
+                ch.on_request('exit-status') do |ch, data|
+                  code = data.read_long
+                  if code == 0
+                    send_start.call
```
adamruzicka:
```ruby
send_start.call if code.zero?
```

lib/smart_proxy_remote_execution_ssh/api.rb
```diff
@@ -4,6 +4,120 @@ class Api < ::Sinatra::Base
...
+
+                ch.on_extended_data do |ch2, type, data|
+                  err_buf += data
+                end
+              end
+            end
+
+            channel.wait
+            if !started
+              send_error.call(400, err_buf)
```
adamruzicka:
```ruby
send_error.call(400, err_buf) unless started
```

lib/smart_proxy_remote_execution_ssh/api.rb
```diff
@@ -4,6 +4,120 @@ class Api < ::Sinatra::Base
...
+            end
+          end
+        rescue Net::SSH::AuthenticationFailed => e
+          send_error.call(401, e.message)
+        rescue Exception => e
+          # TODO - log the backtrace
+          send_error.call(501, "Internal error")
+        end
+        if not socket.closed?
```
adamruzicka:
```ruby
unless socket.closed?
```

lib/smart_proxy_remote_execution_ssh/webrick_ext.rb
```diff
@@ -0,0 +1,11 @@
+module SmartProxyRemoteExecutionSsh
+  module WEBrickExt
+    # An extension to ::WEBrick::HTTPRequest to expost the socket object for highjacking for cockpit
```
iNecas: @mvollmer this set of changes should make the code compatible with webrick as well (at least during my tests everything worked)

lib/smart_proxy_remote_execution_ssh/api.rb
```diff
@@ -1,9 +1,133 @@
...
   module Ssh
     class Api < ::Sinatra::Base
+      include Sinatra::Authorization::Helpers
+
       get "/pubkey" do
         File.read(Ssh.public_key_file)
       end
+
+      post "/session" do
+        do_authorize
```
iNecas: @mvollmer I've noticed this endpoint was not properly authorized (I'm glad we've noticed this before releasing: saved use some troubles:).

It depends on https://github.com/theforeman/smart-proxy/pull/639 getting in first, so that we can actually use the authorization per-endpoint.

lib/smart_proxy_remote_execution_ssh/webrick_ext.rb
mvollmer: I get this error with a prdocution setup on Centos 7:

```
/usr/share/gems/gems/smart_proxy_remote_execution_ssh-0.2.0/lib/smart_proxy_remote_execution_ssh/webrick_ext.rb:10:in `<module:SmartProxyRemoteExecutionSsh>': private method `prepend' called for WEBrick::HTTPRequest:Class (NoMethodError)
```

lib/smart_proxy_remote_execution_ssh/api.rb
```diff
@@ -28,7 +28,9 @@ class Api < ::Sinatra::Base
         ssh_options[:number_of_password_prompts] = 1
 
         socket = nil
-        if env['rack.hijack?']
+        if env['WEBRICK_SOCKET']
+          socket = env['WEBRICK_SOCKET']
```
mvollmer: Is that enough?  Wont Webrick still try to use that socket and send some headers?  The socket should be closed at that point and maybe that is indeed good enough.

iNecas: Have not found a failing situation yet.

lib/smart_proxy_remote_execution_ssh/api.rb
```diff
@@ -56,7 +56,10 @@ class Api < ::Sinatra::Base
         if env['WEBRICK_SOCKET']
           socket = env['WEBRICK_SOCKET']
         elsif env['rack.hijack?']
-          env['rack.hijack'].call
+          begin
```
iNecas: Once we have the webrick extension, we should not get into this situation.

lib/smart_proxy_remote_execution_ssh/webrick_ext.rb
iNecas: I've pushed a new commit that should fix that

lib/smart_proxy_remote_execution_ssh/api.rb
mvollmer: I get 
```
NameError: undefined local variable or method `do_authorize' for #<Proxy::RemoteExecution::Ssh::Api:0x007ffa4c3c70f0>
/usr/share/gems/gems/smart_proxy_remote_execution_ssh-0.2.0/lib/smart_proxy_remote_execution_ssh/api.rb:34:in `block in <class:Api>'
```

Indeed, I can't find `do_authorize` in https://github.com/theforeman/smart-proxy/pull/639

iNecas: It has been updated as part of the review: now updated here as well

lib/smart_proxy_remote_execution_ssh/api.rb
```diff
@@ -1,9 +1,208 @@
+require 'net/ssh'
+
+# When hijacking the socket of a TLS connection, we get a
+# OpenSSL::SSL::SSLSocket or Puma::MiniSSL::Socket, which don't behave
+# the same as real IO::Sockets.  We need to add recv and send for the
+# benefit of the Net::SSH::BufferedIo mixin, and closed? for our own
+# convenience.
+
+module Puma
+  module MiniSSL
```
iNecas: @mvollmer can you re-verify that this is working still for you? On my devel setup, with puma, I actually don't get to the cockpit instance with this (I was able to get there very occasionally). Webrick ssl or puma non-ssl works well, but the puma ssl seems like having some issues still.

mvollmer: I haven't tested with Puma since a couple of weeks or so...

lib/smart_proxy_remote_execution_ssh/api.rb
```diff
@@ -1,9 +1,251 @@
...
+# When hijacking the socket of a TLS connection, we get a
+# OpenSSL::SSL::SSLSocket or Puma::MiniSSL::Socket, which don't behave
+# the same as real IO::Sockets.  We need to add recv and send for the
+# benefit of the Net::SSH::BufferedIo mixin, and closed? for our own
+# convenience.
+
+
 module Proxy::RemoteExecution
   module Ssh
+    class BufferedSocket
```
iNecas: @mvollmer I've pushed the commit that replaces the monkey-patching with wrapper classes. This gives me more confidence about this code not affecting anything else

lib/smart_proxy_remote_execution_ssh/api.rb
```diff
@@ -1,73 +1,116 @@
...
+      def_delegators(:@socket, :to_io, :addr, :peeraddr, :setsockopt, :getsockopt, :fcntl, :close, :closed?, :do_not_reverse_lookup=)
+
+      def initialize(socket)
+        @socket = socket
+        initialize_buffered_io
       end
-      def recv(n)
-        readpartial(n)
+
+      def resv
```
mvollmer: `recv`?

lib/smart_proxy_remote_execution_ssh/api.rb
```diff
@@ -1,73 +1,116 @@
...
           retry
         end
       end
     end
-  end
-end
 
-module Proxy::RemoteExecution
-  module Ssh
+    class PumaBufferedSocket < BufferedSocket
```
mvollmer: As discussed, I would drop all pretends of supporting Puma in this PR.

lib/smart_proxy_remote_execution_ssh/api.rb
iNecas: For this particular case, I've tried the modifier way and it looks a bit weird. The standard way reads a bit better here

lib/smart_proxy_remote_execution_ssh/api.rb
iNecas: ditto

lib/smart_proxy_remote_execution_ssh/api.rb
iNecas: I would vote for this one to also keep in this form

lib/smart_proxy_remote_execution_ssh/api.rb
iNecas: Updated

lib/smart_proxy_remote_execution_ssh/api.rb
iNecas: updated

lib/smart_proxy_remote_execution_ssh/api.rb
mvollmer: I have tested it in my three-machine setup, and didn't find any issues.  The case where we need to wait in order to avoid returning "" to Net::SSH still happens and is still handled correctly.

iNecas:
ok to test

lib/smart_proxy_remote_execution_ssh/api.rb
```diff
@@ -1,9 +1,234 @@
...
+                end
+
+                ch.on_request('exit-status') do |ch, data|
+                  code = data.read_long
+                  send_start.call if code.zero?
+                  err_buf += "Process exited with code #{code}.\r\n"
+                  ch.close
+                end
+
+                channel.on_request('exit-signal') do |ch, data|
```
iNecas: @mvollmer is there a reason why this is using `channel` while the rest is using `ch`?

iNecas:
@mvollmer I've done some refactoring, mainly moving the code around into separate class and splitting it into smaller methods. Waiting for feedback + additional testing (everything worked for me)

lib/smart_proxy_remote_execution_ssh/cockpit.rb
```diff
@@ -0,0 +1,252 @@
+require 'net/ssh'
+require 'forwardable'
+
+module Proxy::RemoteExecution
+  module Cockpit
+    # A wrapper class around different kind o sockets to comply with Net::SSH event loop
```
mvollmer: "different kind o**f** sockets"

smart_proxy_remote_execution_ssh.gemspec
```diff
@@ -31,4 +31,5 @@ Gem::Specification.new do |gem|
   gem.add_development_dependency('rubocop', '0.32.1')
 
   gem.add_runtime_dependency('smart_proxy_dynflow', '>= 0.1.0', '< 0.3.0')
+  gem.add_runtime_dependency('net-ssh')
```
mvollmer: I had to use version 4.2 of `net_ssh` for some compatibility reason... I can try to dog out the reason.

lib/smart_proxy_remote_execution_ssh/cockpit.rb
```diff
@@ -0,0 +1,252 @@
...
+
+    class Session
+      include ::Proxy::Log
+
+      def initialize(env)
+        @env = env
+      end
+
+      def valid?
+        @env["HTTP_CONNECTION"] == "upgrade" && @env["HTTP_UPGRADE"] == "raw"
```
mvollmer: `HTTP_UPGRADE` can be a list, and it is enough if "raw" is in that list.  This allows for future extensions to other protocols.

lib/smart_proxy_remote_execution_ssh/api.rb
mvollmer: I am not aware of any reason, this probably comes from some example code that I used to get started.

lib/smart_proxy_remote_execution_ssh/cockpit.rb
iNecas: Updated

lib/smart_proxy_remote_execution_ssh/cockpit.rb
iNecas: Updated

smart_proxy_remote_execution_ssh.gemspec
iNecas: I was testing with net-ssh 5.2.0 in my devel setup and everything seem to work (except Puma - I wonder if that could be related)

iNecas:
Rebased and merged: going to release a new version of smart-proxy-rex now.

PR #43, opened 2019-05-17
ssh/session: Support for Puma::MiniSSL::Socket

mvollmer:


mvollmer:
This "works", but can be very slow for some reason.  More debugging needed.

mvollmer:
> This "works", but can be very slow for some reason. More debugging needed.

Aha, turns out we can re-use the code we already had for OpenSSL::SSL::SSLSocket.

lib/smart_proxy_remote_execution_ssh/cockpit.rb
```diff
@@ -45,9 +45,9 @@ def self.applies_for?(socket)
       end
     end
 
-    class OpenSSLBufferedSocket < BufferedSocket
+    class SSLBufferedSocket < BufferedSocket
       def self.applies_for?(socket)
-        socket.is_a? ::OpenSSL::SSL::SSLSocket
+        socket.is_a? ::OpenSSL::SSL::SSLSocket or socket.is_a? ::Puma::MiniSSL::Socket
```
iNecas: Not tested yet, but just in case Puma is not installed:

```ruby
return true if socket.is_a? ::OpenSSL::SSL::SSLSocket 
return true if defined?  ::Puma::MiniSSL::Socket && socket.is_a? ::Puma::MiniSSL::Socket
```

iNecas:
I still have no luck with getting this work, currently I see this on proxy during the upgrade phase. 

```
2019-06-10T17:36:56 b5b99b09 [D] /home/inecas/active/projects/cockpit/smart_proxy_remote_execution_ssh/lib/smart_proxy_remote_execution_ssh/cockpit.rb:61:in `+'
/home/inecas/active/projects/cockpit/smart_proxy_remote_execution_ssh/lib/smart_proxy_remote_execution_ssh/cockpit.rb:61:in `recv'
/home/inecas/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/net-ssh-5.2.0/lib/net/ssh/buffered_io.rb:66:in `fill'
/home/inecas/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/net-ssh-5.2.0/lib/net/ssh/connection/session.rb:275:in `block in ev_do_handle_events'
/home/inecas/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/net-ssh-5.2.0/lib/net/ssh/connection/session.rb:271:in `each'
/home/inecas/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/net-ssh-5.2.0/lib/net/ssh/connection/session.rb:271:in `ev_do_handle_events'
/home/inecas/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/net-ssh-5.2.0/lib/net/ssh/connection/event_loop.rb:112:in `ev_select_and_postprocess'
/home/inecas/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/net-ssh-5.2.0/lib/net/ssh/connection/event_loop.rb:31:in `process'
/home/inecas/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/net-ssh-5.2.0/lib/net/ssh/connection/session.rb:228:in `process'
/home/inecas/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/net-ssh-5.2.0/lib/net/ssh/connection/session.rb:181:in `block in loop'
/home/inecas/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/net-ssh-5.2.0/lib/net/ssh/connection/session.rb:181:in `loop'
/home/inecas/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/net-ssh-5.2.0/lib/net/ssh/connection/session.rb:181:in `loop'
/home/inecas/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/net-ssh-5.2.0/lib/net/ssh/connection/session.rb:127:in `close'
/home/inecas/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/net-ssh-5.2.0/lib/net/ssh.rb:256:in `ensure in start'
```

I've tried this, but it just lead for the process to get stuck.

```patch
+++ b/lib/smart_proxy_remote_execution_ssh/cockpit.rb
@@ -58,7 +58,7 @@ module Proxy::RemoteExecution
           # loop, we need to repeatedly call read_nonblock; a single
           # call is not enough.
           while true
-            res += @socket.read_nonblock(n)
+            res += (@socket.read_nonblock(n) || "")
           end
         rescue IO::WaitReadable
           # Sometimes there is no payload after reading everything
```

mvollmer:
> I still have no luck with getting this work,

It works for me in the setup described here: https://github.com/mvollmer/foreman-cockpit
How does your setup differ from that?

Can we see from the Puma source code in what cases `read_nonblock` will return `nil`?

iNecas:
It has been just with development setup, running with latest puma and net::ssh. Do we actually need to wrap the `recv` call in Puma? My understanding was there was something in Webrick ssl that needed that, but not sure if that applies for Puma as well?

mvollmer:
> It has been just with development setup, running with latest puma and net::ssh.

I see, so that means puma 4.0, which means I can't use CentOS 7, right?  I don't really want to setup something custom at this point...

mvollmer:
> Can we see from the Puma source code in what cases read_nonblock will return nil?

When read_nonblock returns `nil` it means EOF.  (It should only do that when given `expection: false`, I guess, and should raise EOFError otherwise, but here we are.)

Does this make sense?  If not, and you don't expect any EOF, then maybe something else is closing the socket down prematurely, and running into `nil` here is just a unexpected symptom of that.

mvollmer:
> Do we actually need to wrap the recv call in Puma?

Yes, since there is no `recv` method in the MiniSSL:Socket class, but it can probably indeed just be a straight call to read_nonblock.  This is what I started out with way back when.

mvollmer:
Just found this, sorry for the "stream of consciousness" comment style: 

https://github.com/puma/puma/commit/acb709b7b74ba7f8129d0fa7ffe4f04e6da24aa2#diff-d8a3dc2e04ffb9e96d5ec39008498878

mvollmer:
And doesn't MiniSSL::Socket::read_nonblock actually always block until there is something to return?

mvollmer:
Okay, I have changed the code to directly use read_nonblock and write_nonblock for MiniSSL.  Everything still works in my CentOS 7 / Foreman 1.22 / Ruby 2.0 / Puma 3.10 setup.

mvollmer:
@iNecas, please have another look.

PR #44, opened 2019-09-30
Update to bundler.d to match other smart proxy plugins

ehelms:


ezr-ondrej:

adamruzicka:
I'm still not convinced this is a good idea. Usually for a project X `bundler.d` and files inside it allow declaring additional gems as dependencies of project X. Since this is a plugin, those two files are used to declare this plugin and its dependency as dependencies for the smart proxy. I'd say it would make things even more confusing if we moved everything to bundler.d

Furthermore, this would declare `smart_proxy_remote_execution_ssh` as dependency of `smart_proxy_remote_execution_ssh` and this plugin only looks at `Gemfile.local.rb`, not at contents of `bundler.d`

adamruzicka:
Now's the time to do this. @ehelms could you drop `bundler.d/foreman_remote_execution_core.rb` while you're at it?

ehelms:
Done

adamruzicka (approved):

adamruzicka:
Thank you @ehelms!

PR #46, opened 2020-04-16
Allow newer bundler versions

mmoll:


adamruzicka:
Thank you @mmoll and sorry for the delay

PR #47, opened 2020-04-23
Unpin dependencies

ekohl:
Now that the Smart proxy can use a modern Ruby version and drop < 2.4, it's no longer needed to pin these dependencies. For Rake 10 there are known security issues.


ekohl:
Looks like this will also need a rubocop change and I don't have time for that.

PR #51, opened 2020-11-09
Loosen deps on smart_proxy_dynflow

adamruzicka:


ares (approved):
Thanks @adamruzicka, merging!

PR #52, opened 2021-05-06
Fixes #32543 - Import foreman_remote_execution_core

adamruzicka:


lib/smart_proxy_remote_execution_ssh/plugin.rb
```diff
@@ -9,25 +11,39 @@ class Plugin < Proxy::Plugin
                      :remote_working_dir      => '/var/tmp',
                      :local_working_dir       => '/var/tmp',
                      :kerberos_auth           => false,
-                     :async_ssh               => false
+                     :async_ssh               => false,
+                     # When set to nil, makes REX use the runner's default interval
+                     # :runner_refresh_interval => nil,
+                     :ssh_log_level           => :fatal,
+                     :cleanup_working_dirs    => true
```
ezr-ondrej, line 18: This setting is missing from the example setting file :)

ezr-ondrej (approved):
Just a real minor find, works as expected :+1:
Do we want to bump the minor version so we have that ready?
I think this is already enough to bump to 0.4 and we require it in the REX PR.

ezr-ondrej:

ezr-ondrej (approved):
Thanks @adamruzicka :+1:

PR #53, opened 2021-05-31
Fixes #31516 - Make cockpit work with passphrase

adamruzicka:
protected keys.


ezr-ondrej (approved):
Thanks @adamruzicka ! :+1:

PR #54, opened 2021-06-10
Fixes #32774 - use directly smart_proxy_dynflow

ezr-ondrej:
The ForemanTasksCore has been discontinued and has been moved to
smart_proxy_dynflow. We should use those classes directly and not rely
on the compatibility layer tasks_core gem.

:boom:

### DIFF
--- a/lib/smart_proxy_remote_execution_ssh.rb
+++ b/lib/smart_proxy_remote_execution_ssh.rb
@@ -1,8 +1,7 @@
-require 'foreman_tasks_core'
-require 'smart_proxy_remote_execution_ssh/version'
 require 'smart_proxy_dynflow'
-require 'smart_proxy_remote_execution_ssh/webrick_ext'
+require 'smart_proxy_remote_execution_ssh/version'
 require 'smart_proxy_remote_execution_ssh/plugin'
+require 'smart_proxy_remote_execution_ssh/webrick_ext'
 
 module Proxy::RemoteExecution
   module Ssh
@@ -53,8 +52,5 @@ module Proxy::RemoteExecution
         Plugin.settings.ssh_log_level = Plugin.settings.ssh_log_level.to_sym
       end
     end
-
-    require 'smart_proxy_dynflow_core/task_launcher_registry'
-    SmartProxyDynflowCore::TaskLauncherRegistry.register('ssh', ForemanTasksCore::TaskLauncher::Batch)
   end
 end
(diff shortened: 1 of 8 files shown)


lib/smart_proxy_remote_execution_ssh.rb
```diff
@@ -4,6 +4,8 @@
 require 'smart_proxy_remote_execution_ssh/webrick_ext'
 require 'smart_proxy_remote_execution_ssh/plugin'
 
+Proxy::Dynflow::TaskLauncherRegistry.register('ssh', Proxy::Dynflow::TaskLauncher::Batch)
```
adamruzicka, line 7: Maybe doing this from the plugin's after activation hook would be better?

adamruzicka (commented):
Works nicely, left one suggestion

lib/smart_proxy_remote_execution_ssh.rb
ezr-ondrej: Done, just heads up, I've removed two more `require '*_core'`

smart_proxy_remote_execution_ssh.gemspec
```diff
@@ -29,7 +29,6 @@ Gem::Specification.new do |gem|
   gem.add_development_dependency('rack-test', '~> 0')
   gem.add_development_dependency('rubocop', '~> 0.82.0')
 
-  gem.add_runtime_dependency('foreman-tasks-core', '>= 0.3.1')
-  gem.add_runtime_dependency('smart_proxy_dynflow', '~> 0.1')
+  gem.add_runtime_dependency('smart_proxy_dynflow', '~> 0.5')
```
ezr-ondrej, line 32: I'm confused with versioning, is the current 0.4 sufficient, or are we releasing 0.5 for that?

adamruzicka: 0.4 is sp dynflow core to sp dynflow, 0.5 will be tasks core -> sp dynflow. Getting 0.4 through packaging is taking a bit longer than I expected, I'd like to phase out the releases a bit so we don't drop a bunch of releases changing everything at the same time

adamruzicka:
smart_proxy_dynflow-0.5.0 is out

ezr-ondrej:
Needs https://github.com/theforeman/smart_proxy_dynflow/pull/83 and probably another `smart_proxy_dynflow` release

adamruzicka:
smart_proxy_dynflow-0.5.1 is out

adamruzicka (approved):

adamruzicka:
Thank you @ezr-ondrej !

PR #55, opened 2021-06-15
Fixes #30277 - Allow spaces in remote_working_dir

pkoprda:


lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
```diff
@@ -170,8 +170,10 @@ def initialization_script
       su_method = @user_method.instance_of?(SuUserMethod)
       # pipe the output to tee while capturing the exit code in a file
       <<-SCRIPT.gsub(/^\s+\| /, '')
-      | sh -c "(#{@user_method.cli_command_prefix}#{su_method ? "'#{@remote_script} < /dev/null '" : "#{@remote_script} < /dev/null"}; echo \\$?>#{@exit_code_path}) | /usr/bin/tee #{@output_path}
-      | exit \\$(cat #{@exit_code_path})"
+      | sh <<EOF | /usr/bin/tee #{@output_path}
+      | #{@user_method.cli_command_prefix}#{su_method ? "'#{@remote_script} < /dev/null '" : "#{@remote_script} < /dev/null"}; echo \\$?>#{@exit_code_path};
+      | exit \\$(cat #{@exit_code_path})
+      | EOF
```
adamruzicka, line 176: Maybe just a matter of personal taste, but could you use a squiggly heredoc here instead of a regular heredoc with substitution?

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
```diff
@@ -169,9 +169,11 @@ def prepare_start
     def initialization_script
       su_method = @user_method.instance_of?(SuUserMethod)
       # pipe the output to tee while capturing the exit code in a file
-      <<-SCRIPT.gsub(/^\s+\| /, '')
-      | sh -c "(#{@user_method.cli_command_prefix}#{su_method ? "'#{@remote_script} < /dev/null '" : "#{@remote_script} < /dev/null"}; echo \\$?>#{@exit_code_path}) | /usr/bin/tee #{@output_path}
-      | exit \\$(cat #{@exit_code_path})"
+      <<~SCRIPT
+        sh <<EOF | /usr/bin/tee #{@output_path}
+        #{@user_method.cli_command_prefix}#{su_method ? "'#{@remote_script} < /dev/null '" : "#{@remote_script} < /dev/null"}; echo \\$?>#{@exit_code_path};
```
adamruzicka, line 174: Since this doesn't now have to be on a single line, could you replace the semicolons with newlines?

adamruzicka (changes requested):
Working directory cleanup doesn't seem to work with the polling runner. My guess would be double escaping

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
```diff
@@ -169,9 +169,12 @@ def prepare_start
...
       # pipe the output to tee while capturing the exit code in a file
-      <<-SCRIPT.gsub(/^\s+\| /, '')
-      | sh -c "(#{@user_method.cli_command_prefix}#{su_method ? "'#{@remote_script} < /dev/null '" : "#{@remote_script} < /dev/null"}; echo \\$?>#{@exit_code_path}) | /usr/bin/tee #{@output_path}
-      | exit \\$(cat #{@exit_code_path})"
+      <<~SCRIPT
+        sh <<EOF | /usr/bin/tee #{@output_path}
+        #{@user_method.cli_command_prefix}#{su_method ? "'#{@remote_script} < /dev/null '" : "#{@remote_script} < /dev/null"}
+        echo \\$?>#{@exit_code_path}
+        EOF
+        exit $(cat #{@exit_code_path})
```
adamruzicka, line 177: This is interesting. It seems to break job cancellation in case there are spaces in remote_working_dir.

```
# Before cancel
$ pstree -p 383
ash(383)-+-sh(384)---busybox(386)---sleep(387)
         `-tee(385)

# After cancel
$ pstree -p 383
ash(383)---tee(385)

# After cancel, here we can see sleep(387) being a child of init(1)
$ pstree -p 1
sshd(1)-+-sleep(387)
        `-sshd(378)---ash(383)---tee(385)
```

The shell process gets killed, but whatever was running inside it gets re-parented under init, where it continues to run.

How about instead of killing by name, we would write the parent process ID to a special location and then on cancel we would read the pid and send the signal directly?

adamruzicka (changes requested):

lib/smart_proxy_remote_execution_ssh/runners/polling_script_runner.rb
```diff
@@ -36,12 +36,12 @@ def prepare_start
...
       close_stdin = '</dev/null'
       close_fds = close_stdin + ' >/dev/null 2>/dev/null'
-      main_script = "(#{@remote_script} #{close_stdin} 2>&1; echo $?>#{@base_dir}/init_exit_code) >#{@base_dir}/output"
+      main_script = "(#{@remote_script_wrapper} #{@remote_script} #{close_stdin} 2>&1; echo $?>#{@base_dir}/init_exit_code) >#{@base_dir}/output"
       control_script_finish = "#{@control_script_path} init-script-finish"
       <<-SCRIPT.gsub(/^ +\| /, '')
       | export CONTROL_SCRIPT="#{@control_script_path}"
       | sh -c '#{main_script}; #{control_script_finish}' #{close_fds} &
-      | echo $! > '#{@base_dir}/pid'
+      | echo $! > /dev/null
```
adamruzicka, line 44: Why write it at all? We could just discard this line completely

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
```diff
@@ -178,8 +183,8 @@ def refresh
     end
 
     def kill
-      if @session
-        run_sync("pkill -f #{remote_command_file('script')}")
+      if @pid_path
```
adamruzicka, line 186: `@pid_path` will almost always be set, it is by no means a good representation of whether something runs on the remote end or not. In this case I'd vote for a default implementation using `@session` for the regular script runner and then override it with unconditional remote kill for the polling runner.

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
```diff
@@ -178,8 +183,8 @@ def refresh
...
+        run_sync("pkill -P $(cat #{@pid_path})")
```
adamruzicka, line 187: `pkill` can read pids from a file on its own with `--pidfile`, there's really no need to fork a `cat`

adamruzicka (commented):
In general seems to work well, left couple of nitpicks

lib/smart_proxy_remote_execution_ssh/runners/polling_script_runner.rb
pkoprda: I thought that it wouldn't work without that. But I was wrong, it didn't work because I also removed `&`.

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
adamruzicka: Ah, sorry, this was a bad suggestion. Reading the manpage I got the impression one could use both `-P` and `--pidfile` and make `pkill` kill processes which are children of processes listed in the file. Sadly it doesn't work like that :/

adamruzicka (approved):
Works well, ACK

PR #62, opened 2021-07-13
Fixes #33002 - Do not use deprecated net-ssh options

adamruzicka:
Additionally fixes a similar error with the value.

```
:paranoid is deprecated, please use :verify_host_key. Supported values are exactly the same, only the name of the option has changed.
```

```
verify_host_key: true is deprecated, use :accept_new_or_local_tunnel
```

### DIFF
--- a/lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
+++ b/lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
@@ -102,6 +102,8 @@ module Proxy::RemoteExecution::Ssh::Runners
     EXPECTED_POWER_ACTION_MESSAGES = ['restart host', 'shutdown host'].freeze
     DEFAULT_REFRESH_INTERVAL = 1
     MAX_PROCESS_RETRIES = 4
+    VERIFY_HOST_KEY = Gem::Version.create(Net::SSH::Version::STRING) < Gem::Version.create('5.0.0') ||
+                      :accept_new_or_local_tunnel
 
     def initialize(options, user_method, suspended_action: nil)
       super suspended_action: suspended_action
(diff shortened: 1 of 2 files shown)


ekohl (changes requested):
We should set the minimum version to 4.2.0 since that introduced support for `verify_host_key` (https://github.com/net-ssh/net-ssh/commit/c3fbcdc882f6763fa718f6adbd9d1d60eaf04b62) :
https://github.com/theforeman/smart_proxy_remote_execution_ssh/blob/f961831d999acace38be92822c591f9253619519/smart_proxy_remote_execution_ssh.gemspec#L33

ekohl:
Oh sorry, 5.0.0 since that added support for `accept_new_or_local_tunnel`

adamruzicka:
Then we'll have to live with `verify_host_key: true is deprecated, use :accept_new_or_local_tunnel` in development since we can't really ship anything above 4.2 in production

ekohl (approved):
> Then we'll have to live with `verify_host_key: true is deprecated, use :accept_new_or_local_tunnel` in development since we can't really ship anything above 4.2 in production

Does it make sense to inspect the version? Anyway, :+1: for this since it removes at least one deprecation. It'd be nice if we can also prevent a deprecation showing up on Debian (Buster ships 5.1.0)

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
```diff
@@ -102,6 +102,8 @@ class ScriptRunner < Proxy::Dynflow::Runner::Base
     EXPECTED_POWER_ACTION_MESSAGES = ['restart host', 'shutdown host'].freeze
     DEFAULT_REFRESH_INTERVAL = 1
     MAX_PROCESS_RETRIES = 4
+    VERIFY_HOST_KEY = Gem.loaded_specs['net-ssh'].version < Gem::Version.create('5.0.0') ||
```
ekohl, line 105: Can `Gem.loaded_specs` be empty? I'm not sure about production RPM deployments.

adamruzicka: Not sure, changed.

ekohl (approved):

ekohl:
Do you want to squash the commits?

adamruzicka:
Squashed

ekohl (approved):

PR #63, opened 2021-08-02
Fixes #33198 - Drop net/ssh in favor of using system ssh binary

pkoprda:


lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
```diff
@@ -329,32 +330,28 @@ def run_sync(command, stdin = nil)
...
+
+      unless stdin.nil?
+        write.puts(stdin)
+        write.close
+        loop do
+          ret = begin
+                  master.gets
+                rescue Errno::EIO
+                  nil
+                end
```
adamruzicka, line 348: The exception gets raised only when you attempt to read after the source has reached eof. Since you're looping anyway, you could start checking for `#eof?` instead of catching the exception.

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
```diff
@@ -329,32 +330,28 @@ def run_sync(command, stdin = nil)
...
-          raise 'could not execute command' unless success
-
-          started = true
+      master, slave = PTY.open
+      read, write = IO.pipe
+      pid = spawn('/usr/bin/ssh', @ssh_user + '@' + @host, command, :in => read, :out => slave)
+      read.close
+      slave.close
+
+      unless stdin.nil?
```
adamruzicka, line 340: Do I read it right that if there's no input, it does not read the output of the command being run?

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
```diff
@@ -329,32 +330,28 @@ def run_sync(command, stdin = nil)
...
-          ch.close
-          ch.wait
-        end
-        ch.exec command do |_, success|
-          raise 'could not execute command' unless success
-
-          started = true
+      master, slave = PTY.open
+      read, write = IO.pipe
+      pid = spawn('/usr/bin/ssh', @ssh_user + '@' + @host, command, :in => read, :out => slave)
```
adamruzicka, line 336: The old implementation was setting a couple of options when opening the connection https://github.com/theforeman/smart_proxy_remote_execution_ssh/blob/master/lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb#L271 . We need to do something similar here. Some of them have direct counterpart in ssh's options and we should use those. Take a look at `man 1 ssh`, search for `-o option`

adamruzicka (commented):
Solid start, let's continue with it

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
pkoprda: Yes, because we checked if input is not empty when we used Net::SSH module:
```ruby
channel.send_data(stdin) unless stdin.nil?
```

adamruzicka: Well yes, but originally if the input was empty, we didn't send it over but we still read the output. Currently it seems that if there's no input, we discard all the output that the remote command produces

pkoprda: Oh okay, I understand now, thanks for the notice :+1:

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
pkoprda: I assume that option `ssh_options[:paranoid] = true` is deprecated. But what about this option:
```ruby
ssh_options[:logger] = Proxy::RemoteExecution::Ssh::LogFilter.new(Proxy::Dynflow::Log.instance)
```
How can be this option included here?

adamruzicka: > I assume that option ssh_options[:paranoid] = true is deprecated. 

It is "deprecated", it just got renamed to something else in later net-ssh versions. I think we want something like `StrictHostKeyChecking=accept-new` would be a fitting equivalent.

> ssh_options[:logger]

We can't really do anything about this one, but it was always more of an workaround so no action is needed here.

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
```diff
@@ -329,32 +330,36 @@ def run_sync(command, stdin = nil)
...
-        ch.exec command do |_, success|
-          raise 'could not execute command' unless success
-
-          started = true
-        end
+      master, slave = PTY.open
+      read, write = IO.pipe
+
+      pid = spawn("/usr/bin/sshpass", "-p #{ssh_options[:password]}", "-P #{ssh_options[:passphrase]}",
+                  "/usr/bin/ssh", "#{@ssh_user}@#{@host}",
```
adamruzicka, line 338: Additionally we can move the user out into `-o User=#{@ssh_user}`

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
```diff
@@ -329,32 +330,36 @@ def run_sync(command, stdin = nil)
...
+                  "-o Port=#{ssh_options[:port].to_s}",
```
adamruzicka, line 339: `.to_s` should not be needed

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
```diff
@@ -329,32 +330,36 @@ def run_sync(command, stdin = nil)
...
+                  "-o IdentityFile=#{ssh_options[:keys][1..-2]}",
```
adamruzicka, line 340: Why `[1..-2]`? If I'm reading `#ssh_options` correctly, there should always be 0 or 1 key

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
```diff
@@ -329,32 +330,36 @@ def run_sync(command, stdin = nil)
...
+                  "-o AddKeysToAgent=#{ssh_options[:keys_only]}",
+                  "-o PasswordAuthentication=#{ssh_options[:auth_methods].include? 'password'}",
+                  "-o PubkeyAuthentication=#{ssh_options[:auth_methods].include? 'publickey'}",
```
adamruzicka, line 343: It would probably more future proof if we replace this with `-o PreferredAuthentications=#{ssh_options[:auth_methods].join(',')` to, among other things, express in which order the methods should be tried.

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
```diff
@@ -329,32 +330,36 @@ def run_sync(command, stdin = nil)
...
+                  "-o UserKnownHostsFile=#{ssh_options[:user_known_hosts_file]}",
```
adamruzicka, line 344: We may need to deal with situation when `ssh_options[:user_known_hosts_file]` is `nil`

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
```diff
@@ -329,32 +330,36 @@ def run_sync(command, stdin = nil)
...
-        end
-        ch.exec command do |_, success|
-          raise 'could not execute command' unless success
-
-          started = true
-        end
+      master, slave = PTY.open
+      read, write = IO.pipe
+
+      pid = spawn("/usr/bin/sshpass", "-p #{ssh_options[:password]}", "-P #{ssh_options[:passphrase]}",
```
adamruzicka, line 337: This works, but then the password could be read by anyone who can list running processes on the smart-proxy machine. sshpass can take the password from environment variable `SSHPASS` which should be a better approach.

adamruzicka, line 337: How about instead of passing the arguments directly, we would first collect everything into an array and then splat the array here. It would allow us to selectively add options more easily.

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
```diff
@@ -329,32 +330,36 @@ def run_sync(command, stdin = nil)
...
+                  "/usr/bin/ssh", "#{@ssh_user}@#{@host}",
+                  "-o Port=#{ssh_options[:port].to_s}",
+                  "-o IdentityFile=#{ssh_options[:keys][1..-2]}",
+                  "-o AddKeysToAgent=#{ssh_options[:keys_only]}",
```
adamruzicka, line 341: Not sure if this is the right option. I'd probably go with `IdentitiesOnly` here

adamruzicka (commented):
Overall step in the right direction. Left some comments inline

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
pkoprda: My bad, now I realized that it's array. I thought that it's string, so I just removed first and last character

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
```diff
@@ -331,20 +331,22 @@ def run_sync(command, stdin = nil)
       stderr = ''
       exit_status = nil
 
+      options = [
+        "-o User=#{@ssh_user}",
+        "-o Port=#{ssh_options[:port]}",
+        "-o IdentityFile=#{ssh_options[:keys][0]}",
```
adamruzicka, line 337: What happens if any of these is unspecified?

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
```diff
@@ -331,20 +331,22 @@ def run_sync(command, stdin = nil)
...
+        "-o IdentitiesOnly=#{ssh_options[:keys_only]}",
+        "-o StrictHostKeyChecking=accept-new",
+        "-o PreferredAuthentications=#{ssh_options[:auth_methods].join(',')}",
+        "-o UserKnownHostsFile=#{ssh_options[:user_known_hosts_file] ? ssh_options[:user_known_hosts_file] : ''}",
+        "-o NumberOfPasswordPrompts=#{ssh_options[:number_of_password_prompts]}",
+        "-o LogLevel=#{ssh_options[:verbose]}",
+      ]
+      ENV['SSHPASS'] = ssh_options[:password]
```
adamruzicka, line 345: We don't want to set this for the entire smart-proxy process. There is a way how to pass a new environment into `spawn`.

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
```diff
@@ -331,20 +331,22 @@ def run_sync(command, stdin = nil)
...
       exit_status = nil
 
+      options = [
+        "-o User=#{@ssh_user}",
+        "-o Port=#{ssh_options[:port]}",
+        "-o IdentityFile=#{ssh_options[:keys][0]}",
+        "-o IdentitiesOnly=#{ssh_options[:keys_only]}",
+        "-o StrictHostKeyChecking=accept-new",
+        "-o PreferredAuthentications=#{ssh_options[:auth_methods].join(',')}",
+        "-o UserKnownHostsFile=#{ssh_options[:user_known_hosts_file] ? ssh_options[:user_known_hosts_file] : ''}",
```
adamruzicka, line 341: It would be better to not pass the option at all if we don't have a value for it

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
adamruzicka: This still applies

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
pkoprda: Is it possible to check values in array or do I have to insert each value separately with `<<` and check that inserted value? For example like this:
```ruby
options << "-o Port=#{port}" if port
options << "-o IdentityFile=#{keys[0]}" if keys
```

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
```diff
@@ -331,22 +331,20 @@ def run_sync(command, stdin = nil)
...
-        "-o UserKnownHostsFile=#{ssh_options[:user_known_hosts_file] ? ssh_options[:user_known_hosts_file] : ''}",
-        "-o NumberOfPasswordPrompts=#{ssh_options[:number_of_password_prompts]}",
-        "-o LogLevel=#{ssh_options[:verbose]}",
-      ]
-      ENV['SSHPASS'] = ssh_options[:password]
+      options = []
+      options << "-o User=#{@ssh_user}"
+      options << "-o Port=#{ssh_options[:port]}" if ssh_options[:port]
+      options << "-o IdentityFile=#{ssh_options[:keys][0]}" if ssh_options[:keys]
+      options << "-o IdentitiesOnly=#{ssh_options[:keys_only]}"
```
adamruzicka, line 338: According to the man page, this option accepts `yes` and `no` values. So far we've never changed the value so we can probably hardcode it here.

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
```diff
@@ -331,22 +331,20 @@ def run_sync(command, stdin = nil)
...
+      options << "-o StrictHostKeyChecking=accept-new"
+      options << "-o PreferredAuthentications=#{ssh_options[:auth_methods].join(',')}"
+      options << "-o UserKnownHostsFile=#{ssh_options[:user_known_hosts_file]}" if ssh_options[:user_known_hosts_file]
+      options << "-o NumberOfPasswordPrompts=#{ssh_options[:number_of_password_prompts]}"
+      options << "-o LogLevel=#{ssh_options[:verbose]}"
 
       master, slave = PTY.open
       read, write = IO.pipe
-      pid = spawn("/usr/bin/sshpass", "-e", "-P #{ssh_options[:passphrase]}", "/usr/bin/ssh", "#{@host}", *options, command, :in => read, :out => slave)
+      pid = spawn({"SSHPASS" => "#{ssh_options[:password]}"}, "/usr/bin/sshpass", "-e", "-P #{ssh_options[:passphrase]}", "/usr/bin/ssh", "#{@host}", *options, command, :in => read, :out => slave)
```
adamruzicka, line 347: Interpolating a string with a string value shouldn't be necessary.

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
```diff
@@ -329,32 +330,36 @@ def run_sync(command, stdin = nil)
...
+      options << "-o IdentitiesOnly=#{ssh_options[:keys_only]}"
+      options << "-o StrictHostKeyChecking=accept-new"
+      options << "-o PreferredAuthentications=#{ssh_options[:auth_methods].join(',')}"
+      options << "-o UserKnownHostsFile=#{ssh_options[:user_known_hosts_file]}" if ssh_options[:user_known_hosts_file]
+      options << "-o NumberOfPasswordPrompts=#{ssh_options[:number_of_password_prompts]}"
+      options << "-o LogLevel=#{ssh_options[:verbose]}"
+
+      master, slave = PTY.open
+      read, write = IO.pipe
+      pid = spawn({"SSHPASS" => "#{ssh_options[:password]}"}, "/usr/bin/sshpass", "-e", "-P #{ssh_options[:passphrase]}", "/usr/bin/ssh", "#{@host}", *options, command, :in => read, :out => slave)
```
adamruzicka, line 347: I don't think we want to set `-P` to the key's passphrase as it does something different.

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
```diff
@@ -329,32 +330,36 @@ def run_sync(command, stdin = nil)
...
+      read.close
+      slave.close
+
+      write.puts(stdin) unless stdin.nil?
+      write.close
+
+      loop do
+        stdout += begin
+                    master.gets
+                  rescue Errno::EIO
```
ares, line 357: when does this happen? if it's "expected" could you add a comment when it happens and why? I assume we could see that if the target closes the connection unexpectedly. It would be great to test this and power off the target VM or unplug the cable etc. We may see other exceptions as well.

adamruzicka: It happens when reading from a file descriptor which is already closed. It won't be needed once https://github.com/theforeman/smart_proxy_remote_execution_ssh/pull/63#discussion_r681489452 is addressed.

The expected connection interruption will need to be handled elsewhere

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
```diff
@@ -329,32 +330,36 @@ def run_sync(command, stdin = nil)
...
+      options << "-o IdentitiesOnly=yes"
+      options << "-o StrictHostKeyChecking=accept-new"
+      options << "-o PreferredAuthentications=#{ssh_options[:auth_methods].join(',')}"
+      options << "-o UserKnownHostsFile=#{ssh_options[:user_known_hosts_file]}" if ssh_options[:user_known_hosts_file]
+      options << "-o NumberOfPasswordPrompts=#{ssh_options[:number_of_password_prompts]}"
+      options << "-o LogLevel=#{ssh_options[:verbose]}"
+
+      master, slave = PTY.open
+      read, write = IO.pipe
+      pid = spawn({"SSHPASS" => ssh_options[:password]}, "/usr/bin/sshpass", "-e", "/usr/bin/ssh", "#{@host}", *options, command, :in => read, :out => slave)
```
ekohl, line 347: Why do you need string interpolation on `@host`?
```suggestion
      pid = spawn({"SSHPASS" => ssh_options[:password]}, "/usr/bin/sshpass", "-e", "/usr/bin/ssh", @host, *options, command, :in => read, :out => slave)
```

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
```diff
@@ -329,32 +330,36 @@ def run_sync(command, stdin = nil)
...
-        end
-        ch.exec command do |_, success|
-          raise 'could not execute command' unless success
-
-          started = true
-        end
+      options = []
+      options << "-o User=#{@ssh_user}"
+      options << "-o Port=#{ssh_options[:port]}" if ssh_options[:port]
+      options << "-o IdentityFile=#{ssh_options[:keys][0]}" if ssh_options[:keys]
```
ekohl, line 337: Should this also check for an empty array?
```suggestion
      options << "-o IdentityFile=#{ssh_options[:keys][0]}" if ssh_options[:keys]&.any?
```

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
```diff
@@ -291,34 +287,10 @@ def settings
...
-        channel.exec(command) do |_, success|
-          @started = true
-          raise('Error initializing command') unless success
-        end
-      end
-      session.process(0) { !run_started? }
+      @command_out, slave = PTY.open
+      @command_in, write = IO.pipe
+      @command_pid = spawn({'SSHPASS' => ssh_options[:password]}, '/usr/bin/sshpass', '-e', '/usr/bin/ssh', @host, *options, command, :in => @command_in, :out => slave)
+      @command_in.close
```
adamruzicka, line 293: This is roughly the same in both `run_sync` and `run_async`. Could we extract it into a method which would open the connection and return all the relevant handles?

adamruzicka (commented):
Currently we're opening a new ssh connection for each call to `run_sync` and `run_async`. For bonus points, it might be worth looking into `ControlMaster`, `ControlPersist` and `ControlPath` ssh options so that we open a multiplexed connection at the beginning and then reuse it for all the things we need to do

pkoprda:
Thank you all for your comments. Please let me know if anything else is needed :)

lib/smart_proxy_remote_execution_ssh/runners/polling_script_runner.rb
```diff
@@ -132,7 +132,7 @@ def cleanup
     def destroy_session
       if @session
         @logger.debug("Closing session with #{@ssh_user}@#{@host}")
-        @session.close
+        system("/usr/bin/ssh", @host, "-o", "ControlPath=/tmp/ssh-#{@ssh_user}@#{@host}", "-O", "exit")
```
ekohl, line 135: Having control paths in a world writable location, especially one that is predictable is insecure. It can be easily exploited. Please don't hardcode it to `/tmp`. At the very least respect the env var `TMPDIR`.

Other than that I have `ControlPath ~/.cache/ssh/%r@%h:%p` in my `~/.ssh/config`. I think you can probably use the same placeholders on the command line. And I wonder if you need to respect ports as well.

One risk I want to warn for is maximum path length. In the past I've had issues with long filenames back when I used a longer static prefix.

I'd recommend a helper to generate the control path given its repetition and potential complexities.

ekohl (changes requested):

adamruzicka:
If I'm not missing anything, the `on_data` method of any of the effective user methods is not being called anywhere so my assumption would be that changing the effective user cannot really work

lib/smart_proxy_remote_execution_ssh/runners/polling_script_runner.rb
```diff
@@ -132,7 +132,7 @@ def cleanup
     def destroy_session
       if @session
         @logger.debug("Closing session with #{@ssh_user}@#{@host}")
-        system("/usr/bin/ssh", @host, "-o", "ControlPath=/tmp/ssh-#{@ssh_user}@#{@host}", "-O", "exit")
+        system("/usr/bin/ssh", @host, "-o", "ControlPath=~/.cache/ssh-#{@ssh_user}@#{@host}", "-O", "exit")
```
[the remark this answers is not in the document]
adamruzicka: As an alternative, the script runner already creates a working directory on the proxy side where it stores its temporary files. Maybe we could put the control socket there as well?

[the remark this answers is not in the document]
pkoprda: I'm not sure if we can store socket on the proxy side, because I think `ControlPath` option can store socket only locally.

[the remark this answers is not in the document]
adamruzicka: Locally is on the proxy side.

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
```diff
@@ -258,32 +257,31 @@ def indent_multiline(string)
...
-      # verify it, otherwise, if missing, import it and continue
-      ssh_options[:verify_host_key] = VERIFY_HOST_KEY
-      ssh_options[:auth_methods] = available_authentication_methods
-      ssh_options[:user_known_hosts_file] = prepare_known_hosts if @host_public_key
-      ssh_options[:number_of_password_prompts] = 1
-      ssh_options[:verbose] = settings[:ssh_log_level]
-      ssh_options[:logger] = Proxy::RemoteExecution::Ssh::LogFilter.new(Proxy::Dynflow::Log.instance)
-      return ssh_options
+      ssh_options = []
+      ssh_options << "-o User=#{@ssh_user}"
```
ekohl, line 273: It surprises me that this works. I would have expected that you would need to use
```suggestion
      ssh_options += ["-o", "User=#{@ssh_user}"]
```

adamruzicka: Yeah, ssh seems to be surprisingly tolerant

pkoprda: It works, because of splat operator (`*`) that I used in `spawn` function.

adamruzicka: That's almost surely not it. Even running `ssh '-o User=foo' target` seems to work

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
```diff
@@ -243,12 +240,14 @@ def close
     rescue StandardError => e
       publish_exception('Error when removing remote working dir', e, false)
     ensure
-      @session.close if @session && !@session.closed?
+      system("/usr/bin/ssh", @host, "-o", "User=#{@ssh_user}", "-o", "ControlPath=#{@local_working_dir}/ssh-%r@%h", "-O", "exit")
```
adamruzicka, line 243: This is also used in the polling runner, could we extract it into a common method?

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
```diff
@@ -258,32 +257,31 @@ def indent_multiline(string)
...
-    def session
-      @session ||= begin
-                     @logger.debug("opening session to #{@ssh_user}@#{@host}")
-                     Net::SSH.start(@host, @ssh_user, ssh_options)
-                   end
+    def session(command)
+      @session = true
+      command_out, slave = PTY.open
+      command_in, write = IO.pipe
+      command_pid = spawn(({'SSHPASS' => @ssh_password} if @ssh_password), '/usr/bin/sshpass', '-e', '/usr/bin/ssh', @host, *ssh_options, command, :in => command_in, :out => slave)
```
adamruzicka, line 267: Thinking about this, we still need a pty, but we need it on the remote side. Could you try experimenting if we could do without dealing with ptys here and just pass `-t` to `ssh`?

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
```diff
@@ -258,32 +257,31 @@ def indent_multiline(string)
...
-      ssh_options[:auth_methods] = available_authentication_methods
-      ssh_options[:user_known_hosts_file] = prepare_known_hosts if @host_public_key
-      ssh_options[:number_of_password_prompts] = 1
-      ssh_options[:verbose] = settings[:ssh_log_level]
-      ssh_options[:logger] = Proxy::RemoteExecution::Ssh::LogFilter.new(Proxy::Dynflow::Log.instance)
-      return ssh_options
+      ssh_options = []
+      ssh_options << "-o User=#{@ssh_user}"
+      ssh_options << "-o Port=#{@ssh_port}" if @ssh_port
+      ssh_options << "-o IdentityFile=#{[@client_private_key_file][0]}" if @client_private_key_file
```
adamruzicka, line 275: Why put the key into an array and then take the first element from the array instead of just passing the key directly?

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
```diff
@@ -258,32 +257,31 @@ def indent_multiline(string)
...
+      ssh_options << "-o IdentitiesOnly=yes"
+      ssh_options << "-o StrictHostKeyChecking=accept-new"
+      ssh_options << "-o PreferredAuthentications=#{available_authentication_methods.join(',')}"
+      ssh_options << "-o UserKnownHostsFile=#{prepare_known_hosts}" if prepare_known_hosts
+      ssh_options << "-o NumberOfPasswordPrompts=1"
+      ssh_options << "-o LogLevel=#{settings[:ssh_log_level]}"
+      ssh_options << "-o ControlMaster=yes"
+      ssh_options << "-o ControlPath=#{@local_working_dir}/ssh-%r@%h"
```
adamruzicka, line 283: Could we extract the actual path into a method? We're constructing it all over the place

lib/smart_proxy_remote_execution_ssh/runners/polling_script_runner.rb
```diff
@@ -132,7 +132,7 @@ def cleanup
     def destroy_session
       if @session
         @logger.debug("Closing session with #{@ssh_user}@#{@host}")
-        @session.close
+        system("/usr/bin/ssh", @host, "-o", "User=#{@ssh_user}", "-o", "ControlPath=#{@local_working_dir}/ssh-%r@%h", "-O", "exit")
```
adamruzicka, line 135: When using `system`, all output the command produces is printed to output streams of the parent process, it should either be silenced or forwarded to proper logs.

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
```diff
@@ -258,32 +257,31 @@ def indent_multiline(string)
...
+      ssh_options << "-o User=#{@ssh_user}"
+      ssh_options << "-o Port=#{@ssh_port}" if @ssh_port
+      ssh_options << "-o IdentityFile=#{[@client_private_key_file][0]}" if @client_private_key_file
+      ssh_options << "-o IdentitiesOnly=yes"
+      ssh_options << "-o StrictHostKeyChecking=accept-new"
+      ssh_options << "-o PreferredAuthentications=#{available_authentication_methods.join(',')}"
+      ssh_options << "-o UserKnownHostsFile=#{prepare_known_hosts}" if prepare_known_hosts
+      ssh_options << "-o NumberOfPasswordPrompts=1"
+      ssh_options << "-o LogLevel=#{settings[:ssh_log_level]}"
+      ssh_options << "-o ControlMaster=yes"
```
adamruzicka, line 282: Setting `ControlMaster` to `yes` sets up the control socket, other connections have to use `ControlMaster` set to `no` to be able to reuse it. Best option here would be setting it to `auto` to perform opportunistic multiplexing

adamruzicka (changes requested):
There seems to be quite significant delay between when the remote process finishes and when the job is actually marked as finished. This doesn't seem to happen for longer running commands, but for commands which exit immediately it is quite noticeable. To make things more interesting, this doesn't seem to happen when there are multiple hosts in the job.

For example if I run `date`, then the date in output was `Fri Sep 10 13:56:48 UTC 2021`, but according to logs the exit request was sent 20 seconds later `Sep 10 13:57:08 foreman-nightly-20210910.example.com smart-proxy[26916]: Exit request sent.`

ekohl:
I do wonder: can one job close the control socket for another job? Let's say have this flow:

Job A for host X starts
Job A opens the control socket
Job B starts
Job B reuses the control socket
Job A completes
Job A closes the control socket

What happens to Job B then?

For reference, on my own machine I have this in `~/.ssh.config`:
```
ControlMaster auto
ControlPath ~/.cache/ssh/%r@%h:%p
ControlPersist 1
```

This means it keeps the control socket open for one second after the last SSH connection ends. You can consider this as well, but there are trade offs.

First of all, having it open for a longer time can be a security problem (which is true in general for control sockets) but given that there is no password on the SSH keys I don't think it really makes a difference for REX.

Another is that if a control socket is open but the connection is dead, commands can wait for a while trying to reuse the control socket.

adamruzicka:
> can one job close the control socket for another job?

Possibly. We already have per-job and per-host local directories, we could just store the control socket there.

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
```diff
@@ -262,25 +266,33 @@ def should_cleanup?
...
-      command_pid = spawn(({'SSHPASS' => @ssh_password} if @ssh_password), '/usr/bin/sshpass', '-e', '/usr/bin/ssh', @host, *ssh_options, command, :in => command_in, :out => slave)
-      return command_in, command_out, command_pid, slave, write
+
+      in_read, in_write = IO.pipe
+      out_read, out_write = IO.pipe
+      command_pid = if @ssh_password 
+                      spawn({'SSHPASS' => @ssh_password}, '/usr/bin/sshpass', '-e', '/usr/bin/ssh', @host, *ssh_options, command, :in => in_read, :out => out_write)
+                    else
+                      spawn('/usr/bin/ssh', @host, *ssh_options, command, :in => in_read, :out => out_write)
+                    end
```
adamruzicka, line 276: We can leverage splatting to reduce repetition


```suggestion
      args = []
      args += [{'SSHPASS' => @ssh_password}, '/usr/bin/sshpass', '-e']) if @ssh_password
      args += ['/usr/bin/ssh', @host, ssh_options, command].flatten
      command_pid = spawn(*args, :in => in_read, :out => out_write)
```

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
```diff
@@ -1,5 +1,7 @@
 require 'net/ssh'
 require 'fileutils'
+require 'pty'
```
adamruzicka, line 3: This shouldn't be needed anymore

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
```diff
@@ -258,32 +261,38 @@ def indent_multiline(string)
...
-                   end
+    def session(command)
+      @session = true
+
+      in_read, in_write = IO.pipe
+      out_read, out_write = IO.pipe
+      args = []
+      args += [{'SSHPASS' => @ssh_password}, '/usr/bin/sshpass', '-e'] if @ssh_password
+      args += ['/usr/bin/ssh', @host, ssh_options, command].flatten
+      command_pid = spawn(*args, :in => in_read, :out => out_write)
```
adamruzicka, line 275: Because `:err` is not redirected, ssh debug logging leaks to the parent process' stderr. This should be captured and passed to he job's output

adamruzicka (changes requested):

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
pkoprda: I am not sure if it is now captured properly, because when I used async ssh, stdout is sometimes mixed with stderr

adamruzicka (changes requested):
`sudo` with password doesn't work

```
sudo: no tty present and no askpass program specified
```

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
```diff
@@ -258,32 +230,38 @@ def indent_multiline(string)
...
+      in_read, in_write = IO.pipe
+      out_read, out_write = IO.pipe
+      args = []
+      args += [{'SSHPASS' => @ssh_password}, '/usr/bin/sshpass', '-e'] if @ssh_password
+      args += ['/usr/bin/ssh', @host, ssh_options, command].flatten
+      command_pid = spawn(*args, :in => in_read, [:out, :err] => out_write)
+      in_read.close
+      out_write.close
+
+      return in_write, out_read, command_pid
```
ekohl, line 248: I'm wondering if you really need two pipes. If one is only used for reads and the other only for writes, can't it use a single pipe?

I also wonder if this use case isn't really what [Open3.popen2e](https://ruby-doc.org/stdlib-2.4.1/libdoc/open3/rdoc/Open3.html#method-c-popen2e) is supposed to cover. Wouldn't that be a much better API to use?

adamruzicka: > I'm wondering if you really need two pipes.

I think so. As far as I know we cannot make a pipe, pass both ends to the child process, but also keep them and read/write from/to the ends we kept. Or am I missing something?

adamruzicka (changes requested):
Also there is some net/ssh/krb stuff https://github.com/theforeman/smart_proxy_remote_execution_ssh/pull/63/files#diff-f7a31028d1cbfee3429cdcb724c2fd35effcf3b665e39aec9c348ee015415313R5 we should get rid off and then unconditionally add kerberos to the list of authetnication methods.

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
ekohl: But it immediately closes `in_read` and `out_write`. However, I think popen2e opens the pipes for you so that's probably a lot safer.

ekohl: Can you elaborate on that? It should create a new process in a new thread. You have a `wait_thr` variable (which is an instance of [Process::Status](https://ruby-doc.org/core/Process/Status.html)). You can use `wait_thr.value` and that should block until the process has exited. At least if I read the docs.

pkoprda: The problem is that `Process.wait(@command_pid)` is waiting for the process with pid which doesn't exist, because that process is already dead. It throws this error:
```
error while dispatching request to runner 375588db-51c1-4ec5-84b6-e7c0f225aadf:Errno::ECHILD No child processes
```

ekohl: That's a good point, but I think it would be worth supporting a `Process::Status` instance instead of a PID in Dynflow. @adamruzicka?

ekohl: After some out of band communication I think the current implementation can be the right one. However, I do think it would be worth adding a comment that the method is like `Open3.popen2e` but without a separate thread to monitor it. Otherwise some may come in and try to refactor it again.

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
```diff
@@ -258,32 +222,41 @@ def indent_multiline(string)
...
     end
 
-    def session
-      @session ||= begin
-                     @logger.debug("opening session to #{@ssh_user}@#{@host}")
-                     Net::SSH.start(@host, @ssh_user, ssh_options)
-                   end
+    # Creates session with two pipes - one for reading and one for
+    # writing. Similar to `Open3.popen2` method but without creating
+    # separate thread to monitor it.
```
ekohl, line 230:
```suggestion
    # a separate thread to monitor it.
```

ekohl (commented):
Overall I think the design looks good. Some small details inline.

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
```diff
@@ -210,45 +196,23 @@ def timeout_interval
...
-    rescue IOError, Net::SSH::Disconnect => e
-      @session.shutdown!
-      check_expecting_disconnect
-      if @expecting_disconnect
-        publish_exit_status(0)
-      else
-        publish_exception('Unexpected disconnect', e)
-      end
+    def close_session
+      system("/usr/bin/ssh", @host, "-o", "User=#{@ssh_user}", "-o", "ControlPath=#{control_socket_file}", "-O", "exit", [:out, :err] => "/dev/null")
```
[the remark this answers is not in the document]
pkoprda: Added raising an exception and I will add logging support in the next iteration.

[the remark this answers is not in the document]
pkoprda: Yeah that should be before that raise :+1: 

I wonder if logging stderr with pipes is the best solution. Is there some better solution for this or should I keep it as it is now?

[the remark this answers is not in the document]
adamruzicka: As far as I know we don't really have any alternatives apart from letting it escape to parent process' std{out,err}.

However we could make it a bit cleaner. How about adding something like `run_sync_locally` which would run a command locally, capturing stdout and stderr and then making `run_sync` only generate a command and pass it to `run_sync_locally`?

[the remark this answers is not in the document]
pkoprda: Hmmm I don't know, it sounds a little bit unnecessary. Or maybe I don't understand how it makes the code cleaner. Did you mean it like `run_sync` would looks like this:
```ruby
def run_sync(command)
  return session(command)
end
```
and I would call this method from `run_sync_locally`?

pkoprda:
@adamruzicka @ekohl thank you for your reviews! I just commited some code, can you look at it?
Also what ssh log levels should we support from `man 5 ssh_config`? For example [here](https://github.com/theforeman/smart_proxy_remote_execution_ssh/blob/611f315acbf3a9cf23a4ff500a9ca31fd84a00c4/lib/smart_proxy_remote_execution_ssh/plugin.rb#L3) we support warn log level, but ssh does not support this log level.

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
```diff
@@ -210,45 +196,30 @@ def timeout_interval
...
+    def close_session
+      @session = nil
+      raise "Control socket file does not exist" unless File.exist?(control_socket_file)
+      r_err, w_err = IO.pipe
+      system("/usr/bin/ssh", @host, "-o", "User=#{@ssh_user}", "-o", "ControlPath=#{control_socket_file}", "-O", "exit", :out => "/dev/null", :err => w_err)
+      w_err.close
+      debug_str = ''
+      debug_str += r_err.read until r_err.eof?
+      r_err.close
+      debug_str.lines.each { |line| @logger.debug(line.strip) }
```
ekohl, line 208: Should this be prefixed with some context about where it's from?

ekohl (commented):
>  Also what ssh log levels should we support from `man 5 ssh_config`? For example [here](https://github.com/theforeman/smart_proxy_remote_execution_ssh/blob/611f315acbf3a9cf23a4ff500a9ca31fd84a00c4/lib/smart_proxy_remote_execution_ssh/plugin.rb#L3) we support warn log level, but ssh does not support this log level.

That's a good point and should be supported. I'm not sure what you mean with ssh not supporting it. I can use `ssh -o LogLevel=debug` and I get debug output. Isn't that what the setting controls?

Or do you mean that the names are different? If it's the latter, we can either map values in the code here or modify the config. The installer doesn't even write out the option so we don't really need to worry about compatibility so I'd lean to changing the settings to avoid maintaining a mapping.

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
```diff
@@ -258,38 +226,54 @@ def indent_multiline(string)
...
     end
 
     def settings
       Proxy::RemoteExecution::Ssh::Plugin.settings
     end
 
+    def get_args(command, with_pty = false)
+      args = []
+      args = [{'SSHPASS' => @ssh_password}, '/usr/bin/sshpass', '-e'] if @ssh_password
+      args = [{'SSHPASS' => @key_passphrase}, '/usr/bin/sshpass', '-e'] if @key_passphrase && args.empty?
```
adamruzicka, line 273: This gives us one or the other. I'd expect both to be tried in the order defined by preferred authentication methods

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
```diff
@@ -258,38 +226,54 @@ def indent_multiline(string)
...
     end
 
     def settings
       Proxy::RemoteExecution::Ssh::Plugin.settings
     end
 
+    def get_args(command, with_pty = false)
+      args = []
+      args = [{'SSHPASS' => @ssh_password}, '/usr/bin/sshpass', '-e'] if @ssh_password
+      args = [{'SSHPASS' => @key_passphrase}, '/usr/bin/sshpass', '-e'] if @key_passphrase
```
adamruzicka, line 273: In this case, the later one "wins". This wouldn't work in case the host accepts only the password, but user provided both. One way would be nesting `sshpass`es and having them read the password either from a file (`-f` flag) or a file descriptor (`-d`)

pkoprda: Is it safe to store passwords in a file? In which folder should we store this file?

adamruzicka: That depends, using a file descriptor would definitely be safer. In case we go with a file, the local working directory would be probably the best place.

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
```diff
@@ -258,38 +226,57 @@ def indent_multiline(string)
...
 
     def settings
       Proxy::RemoteExecution::Ssh::Plugin.settings
     end
 
+    def get_args(command, with_pty = false)
+      fd = IO.sysopen(local_command_file("authentication_keys"), 'w+')
+      io = IO.new(fd)
+      io << "#{@ssh_password}\n" if @ssh_password
+      io << @key_passphrase if @key_passphrase
```
adamruzicka, line 274: What happens if both `@ssh_password` and `@key_passphrase` are provided? How does sshpass deal with that?

pkoprda: I thought that file is read line by line with both `@ssh_password` and `@key_passphrase`

adamruzicka: Have you verified that? The manpage doesn't really indicate it would expect multiple passwords to try

adamruzicka: And neither does the source code https://github.com/kevinburke/sshpass/blob/master/main.c#L477

pkoprda: No, I was just thinking that it might work, but it really doesn't

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
```diff
@@ -258,38 +226,61 @@ def indent_multiline(string)
...
 
     def settings
       Proxy::RemoteExecution::Ssh::Plugin.settings
     end
 
+    def create_fd(sshpass)
+      fd = IO.sysopen(local_command_file("authentication_keys"), 'w+')
+      io = IO.new(fd)
+      io << sshpass
+      fd
```
adamruzicka, line 274: In theory this could be an anonymous pipe opened with `IO.pipe`

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
```diff
@@ -258,38 +226,61 @@ def indent_multiline(string)
...
+    end
+
+    def get_args(command, with_pty = false)
+      args = []
+      args += ['/usr/bin/sshpass', '-d', create_fd(@ssh_password).to_s] if @ssh_password
+      args += ['/usr/bin/sshpass', '-d', create_fd(@key_passphrase).to_s] if @key_passphrase
```
adamruzicka, line 280: I don't think this would work. When child sshpass process is spawned, it has only access to the standard input/output/error. If you're passing in a file descriptor, you need to tell `spawn` to explicitly pass the opened file here https://github.com/theforeman/smart_proxy_remote_execution_ssh/pull/63/files#diff-f7a31028d1cbfee3429cdcb724c2fd35effcf3b665e39aec9c348ee015415313R241

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
pkoprda: Hmmm but I don't know how to connect it with file descriptor

adamruzicka: What do you mean by connecting it with file descriptor? `IO.pipe` gives you two IOs, calling `#to_i` on each of them gives you the associated file descriptor.

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
```diff
@@ -258,38 +226,61 @@ def indent_multiline(string)
...
+      io_r, io_w = IO.pipe
+      io_w << sshpass
+      io_w.close
+      io_r
+    end
+
+    def get_args(command, with_pty = false)
+      args = []
+      args += ['/usr/bin/sshpass', '-d', create_fd(@ssh_password).fileno.to_s] if @ssh_password
+      args += ['/usr/bin/sshpass', '-P', 'passphrase', '-d', create_fd(@key_passphrase).fileno.to_s] if @key_passphrase
```
adamruzicka, line 280: I'm afraid we might be leaking FDs here. I don't really see the reading end being closed after the child process is forked. It would be good to have per-run index of all FDs/pipes we've opened so that they can be closed properly

adamruzicka, line 280: In my setup I have a passphrase-less key and a password. When I log in with the key it works, when I log in with the password it works, when I try to log in with the key and a passphrase it works, although the passphrase is never really used.

However when I provide both a passphrase and a password, it gets stuck when opening the initial connection.

```
# pstree -p 2688
sshpass(2688)───sshpass(2689)───ssh(2690)

# ps -f -p 2688 -p 2689 -p 2690 | cat
UID        PID  PPID  C STIME TTY          TIME CMD
foreman+  2688  2531  0 11:54 ?        00:00:00 /usr/bin/sshpass -d 16 /usr/bin/sshpass -P passphrase -d 18 /usr/bin/ssh skilled-foal.example.com -o User=root -o Port=22 -o IdentityFile=/var/lib/foreman-proxy/ssh/id_rsa_foreman_proxy -o IdentitiesOnly=yes -o StrictHostKeyChecking=no -o PreferredAuthentications=password,publickey -o NumberOfPasswordPrompts=1 -o LogLevel=fatal -o ControlMaster=auto -o ControlPath=/var/tmp/foreman-proxy/foreman-ssh-cmd-cea5225c-d810-4349-9c63-38616a32d692/socket -o ControlPersist=yes mkdir -p /var/tmp/foreman-ssh-cmd-cea5225c-d810-4349-9c63-38616a32d692
foreman+  2689  2688  0 11:54 pts/2    00:00:00 /usr/bin/sshpass -P passphrase -d 18 /usr/bin/ssh skilled-foal.example.com -o User=root -o Port=22 -o IdentityFile=/var/lib/foreman-proxy/ssh/id_rsa_foreman_proxy -o IdentitiesOnly=yes -o StrictHostKeyChecking=no -o PreferredAuthentications=password,publickey -o NumberOfPasswordPrompts=1 -o LogLevel=fatal -o ControlMaster=auto -o ControlPath=/var/tmp/foreman-proxy/foreman-ssh-cmd-cea5225c-d810-4349-9c63-38616a32d692/socket -o ControlPersist=yes mkdir -p /var/tmp/foreman-ssh-cmd-cea5225c-d810-4349-9c63-38616a32d692
foreman+  2690  2689  0 11:54 pts/3    00:00:00 /usr/bin/ssh skilled-foal.example.com -o User=root -o Port=22 -o IdentityFile=/var/lib/foreman-proxy/ssh/id_rsa_foreman_proxy -o IdentitiesOnly=yes -o StrictHostKeyChecking=no -o PreferredAuthentications=password,publickey -o NumberOfPasswordPrompts=1 -o LogLevel=fatal -o ControlMaster=auto -o ControlPath=/var/tmp/foreman-proxy/foreman-ssh-cmd-cea5225c-d810-4349-9c63-38616a32d692/socket -o ControlPersist=yes mkdir -p /var/tmp/foreman-ssh-cmd-cea5225c-d810-4349-9c63-38616a32d692

# gdb -p 2688
(gdb) info stack
#0  0x00007effc1de5bbb in pselect () from /lib64/libc.so.6
#1  0x00000000004018f2 in runprogram ()
#2  0x00007effc1d12555 in __libc_start_main () from /lib64/libc.so.6
#3  0x000000000040114d in _start ()

# gdb -p 2689
(gdb) info stack
#0  0x00007f7798a17bbb in pselect () from /lib64/libc.so.6
#1  0x00000000004018f2 in runprogram ()
#2  0x00007f7798944555 in __libc_start_main () from /lib64/libc.so.6
#3  0x000000000040114d in _start ()

# gdb -p 2690
(gdb) info stack
#0  0x00007fb519e24b40 in __read_nocancel () from /lib64/libc.so.6
#1  0x0000555de953b504 in readpassphrase ()
#2  0x0000555de9512eb8 in read_passphrase ()
#3  0x0000555de94e916d in userauth_passwd ()
#4  0x0000555de94ec13b in userauth ()
#5  0x0000555de94ec271 in input_userauth_failure ()
#6  0x0000555de9516ce9 in ssh_dispatch_run ()
#7  0x0000555de94ebb25 in ssh_userauth2 ()
#8  0x0000555de94e69bf in ssh_login ()
#9  0x0000555de94d5bae in main ()
```

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
```diff
@@ -258,38 +226,61 @@ def indent_multiline(string)
...
+    # Creates session with three pipes - one for reading and two for
+    # writing. Similar to `Open3.popen3` method but without creating
+    # a separate thread to monitor it.
+    def session(args, in_stream: true, out_stream: true, err_stream: true)
+      @session = true
+
+      in_read, in_write = in_stream ? IO.pipe : '/dev/null'
+      out_read, out_write = out_stream ? IO.pipe : [nil, '/dev/null']
+      err_read, err_write = err_stream ? IO.pipe : [nil, '/dev/null']
+      command_pid = spawn(*args, :in => in_read, :out => out_write, :err => err_write, :close_others => false)
```
adamruzicka, line 241: I wasn't able to make https://github.com/theforeman/smart_proxy_remote_execution_ssh/pull/63/files#r744680898 work with `:close_others`, but explicitly passing FDs worked

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
```diff
@@ -258,38 +226,54 @@ def indent_multiline(string)
...
     end
 
     def settings
       Proxy::RemoteExecution::Ssh::Plugin.settings
     end
 
+    def get_args(command, with_pty = false)
+      args = []
+      args = [{'SSHPASS' => @ssh_password}, '/usr/bin/sshpass', '-e'] if @ssh_password
+      args = [{'SSHPASS' => @key_passphrase}, '/usr/bin/sshpass', '-e'] if @key_passphrase
```
adamruzicka, line 273: When asking for key passphrase the prompt is different than to the password one so sshpass doesn't match it by default. We need to add '-P' and 'passphrase' to the arguments.

Also the order should be the other way around so password "wins"

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
```diff
@@ -298,67 +282,42 @@ def run_async(command)
...
+
+      if out_io
+        stdout += out_io.read until out_io.eof? rescue
+        out_io.close
+      end
+      debug_str += err_io.read until err_io.eof? rescue
+      err_io.close
+      debug_str.lines.each { |line| @logger.debug(line.strip) }
+
+      return stdout
```
adamruzicka, line 307: please return `debug_str` here as well

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
```diff
@@ -298,67 +282,42 @@ def run_async(command)
...
-      end
-      session.process(0) { !started }
-      # Closing the channel without sending any data gives us SIGPIPE
-      channel.close unless stdin.nil?
-      channel.wait
+
+      pid, tx, rx, err = session(get_args(command))
+      tx.puts(stdin) unless stdin.nil?
+      tx.close
+      stdout = read_output_debug(err, rx)
```
adamruzicka, line 318: capture stdout and stderr returned by read_output_debug, otherwise stderr at L321 will always be an empty string which would make debugging unnecessarily difficult

also you don't need to define stdout, stderr and exit_status on lines 311-313

pkoprda: Yeah, I forgot to remove those definitions. Resolved :+1:

adamruzicka (approved):
Alright, let's get this in.

adamruzicka:
Thank you @pkoprda & @ekohl !

PR #64, opened 2021-08-12
Groundwork for pull provider

adamruzicka:
TODOs:
- [ ] expose this in settings. This will rely on some things pioneered in https://github.com/theforeman/smart_proxy_remote_execution_ssh/pull/57
- [ ] Address todos in code
- [ ] (optional) Add support for token-based auth
- [ ] (optional) Switch polling script runner to the new api


lib/smart_proxy_remote_execution_ssh/actions/pull_script.rb
```diff
@@ -0,0 +1,100 @@
...
+        output[:state] = :delivered
+        suspend
+      else
+        super
+      end
+    end
+
+    def init_run
+      otp_password = if input[:with_mqtt]
+                       ::Proxy::Dynflow::Manager.generate_otp(execution_plan_id)
```
ezr-ondrej, line 26:
```suggestion
                       ::Proxy::Dynflow::OtpManager.generate_otp(execution_plan_id)
```

lib/smart_proxy_remote_execution_ssh/actions/pull_script.rb
```diff
@@ -0,0 +1,100 @@
...
+    end
+
+    def mqtt_notify(payload)
+      MQTT::Client.connect(settings.mqtt_broker, settings.mqtt_port) do |c|
+        c.publish("yggdrasil/#{input[:hostname]}/data/in", JSON.dump(payload), false, 1)
+      end
+    end
+
+    def settings
+      Proxy::Plugin::RemoteExecution::Ssh::Plugin.settings
```
ezr-ondrej, line 93:
```suggestion
      Proxy::RemoteExecution::Ssh::Plugin.settings
```

ezr-ondrej (commented):
We need to work with subscription manager generated UUID instead of hostname, that's where the yggdrassil client listens for mqtt messages :(

adamruzicka:
> We need to work with subscription manager generated UUID instead of hostname

Well, since foreman doesn't know what the proxy side needs, we'll need to send both and let the proxy pick. That means we'll need an extension point on Foreman's side where katello can inject the uuid into the job. Proxy will get a list of "names" and either will pick one depending on the mode, or in case of pull it will make the job available under all of them. How does that sound?

ezr-ondrej:
> Well, since foreman doesn't know what the proxy side needs, we'll need to send both and let the proxy pick. That means we'll need an extension point on Foreman's side where katello can inject the uuid into the job.

Sounds like exactly what we will need.

> in case of pull it will make the job available under all of them

you mean to publish under the topics of the hostname AND uuid? sounds interesting, but probably better safe than sorry

adamruzicka:
> you mean to publish under the topics of the hostname AND uuid?

Probably not publish on mqtt using both, but through the api the job should be obtainable with a cert containing either of those

ezr-ondrej:
Actually the mqtt topic depends on cert CN:
https://github.com/RedHatInsights/yggdrasil/blob/c5dc5ebeb1ec6326a5102cd5460b7209105b968d/cmd/yggd/main.go#L162

lib/smart_proxy_remote_execution_ssh/job_storage.rb
```diff
@@ -0,0 +1,50 @@
...
+                  uuid: uuid,
+                  hostname: hostname,
+                  execution_plan_uuid: execution_plan_uuid,
+                  run_step_id: run_step_id,
+                  job: job)
+      uuid
+    end
+
+    def drop_job(execution_plan_uuid, run_step_id)
+      jobs.where(execution_plan_uuid: execution_plan_id, run_step_id: run_step_id).delete
```
ezr-ondrej, line 37:
```suggestion
      jobs.where(execution_plan_uuid: execution_plan_uuid, run_step_id: run_step_id).delete
```

lib/smart_proxy_remote_execution_ssh/actions/pull_script.rb
```diff
@@ -0,0 +1,100 @@
...
+    def cleanup(_plan = nil)
+      job_storage.drop_job(execution_plan_id, run_step_id)
+      Proxy::Dynflow::OtpManager.passwords.delete(execution_plan_id)
+    end
+
+    def process_external_event(event)
+      output[:state] = :running
+      data = event.data
+      continuous_output = Proxy::Dynflow::ContinuousOutput.new
+      continuous_output.add_output(lines, 'stdout') if data.key?('output')
```
ezr-ondrej, line 43: I'd rather expect multiple lines, but we can send one by one as well I guess.
```suggestion
      Array(data['output']).each { |line| continuous_output.add_output(line, 'stdout') } if data.key?('output')
```

lib/smart_proxy_remote_execution_ssh/actions/pull_script.rb
```diff
@@ -0,0 +1,100 @@
...
+        super
+      end
+    end
+
+    def init_run
+      otp_password = if input[:with_mqtt]
+                       ::Proxy::Dynflow::OtpManager.generate_otp(execution_plan_id)
+                     end
+      input[:job_uuid] = job_storage.store_job(input[:hostname], execution_plan_id, run_step_id, input[:script])
+      output[:state] = :ready_for_pickup
```
ezr-ondrej, line 29:
```suggestion
      output[:state] = :ready_for_pickup
      output[:result] = []
```

ezr-ondrej (commented):
Few more typos and we work if we introduce a hack to check for the uuid as well as for the hostname.

lib/smart_proxy_remote_execution_ssh/actions/pull_script.rb
```diff
@@ -0,0 +1,100 @@
...
+        },
+        content: "#{input[:proxy_url]}/jobs/#{input[:job_uuid]}",
+      }
+      mqtt_notify payload
+      output[:state] = :notified
+    end
+
+    def mqtt_notify(payload)
+      MQTT::Client.connect(settings.mqtt_broker, settings.mqtt_port) do |c|
+        c.publish("yggdrasil/#{input[:hostname]}/data/in", JSON.dump(payload), false, 1)
```
ezr-ondrej, line 88: can we add a function that would give us the topic to be more flexible?
```suggestion
        c.publish(mqtt_host_topic, JSON.dump(payload), false, 1)
```

lib/smart_proxy_remote_execution_ssh/actions/pull_script.rb
```diff
@@ -0,0 +1,100 @@
...
+        type: 'data',
+        message_id: SecureRandom.uuid,
+        version: 1,
+        sent: DateTime.now.iso8601,
+        directive: 'foreman',
+        metadata: {
+          'job_uuid': input[:job_uuid],
+          'username': execution_plan_id,
+          'password': otp_password,
+          'return_url': "#{input[:proxy_url]}/jobs/#{input[:job_uuid]}/update",
```
ezr-ondrej, line 78:
```suggestion
          'return_url': "#{input[:proxy_url]}/ssh/jobs/#{input[:job_uuid]}/update",
```

lib/smart_proxy_remote_execution_ssh/actions/pull_script.rb
```diff
@@ -0,0 +1,100 @@
...
+        },
+        content: "#{input[:proxy_url]}/jobs/#{input[:job_uuid]}",
```
ezr-ondrej, line 80:
```suggestion
        content: "#{input[:proxy_url]}/ssh/jobs/#{input[:job_uuid]}",
```

adamruzicka:
Consumer uuid support needs https://github.com/theforeman/foreman_remote_execution/pull/649 on Foreman's side

lib/smart_proxy_remote_execution_ssh/actions/pull_script.rb
```diff
@@ -0,0 +1,113 @@
...
+        # If the job is not running yet on the client, wipe it from storage
+        cleanup
+        # TODO: Stop the action
+      when :notified, :running
+        # Client was notified or is already running, dealing with this situation
+        # is only supported if mqtt is available
+        # Otherwise we have to wait it out
+        if input[:with_mqtt]
+          cleanup
+          payload = {} # TODO
```
adamruzicka, line 61: @ezr-ondrej What is the expected payload for job cancellation?

ezr-ondrej: It seems to be quite up to us, there is not much documentation on the topic in https://github.com/RedHatInsights/cloud-connector#data tho I'd need to check wether it won't always try to fech content, but probably not if we ommit it.
then I'd go with data message, directive `cancel` ?

lib/smart_proxy_remote_execution_ssh/actions/pull_script.rb
```diff
@@ -0,0 +1,124 @@
...
+    def process_external_event(event)
+      output[:state] = :running
+      data = event.data
+      continuous_output = Proxy::Dynflow::ContinuousOutput.new
+      Array(data['output']).each { |line| continuous_output.add_output(line, 'stdout') } if data.key?('output')
+      exit_code = data['exit_code'].to_i if data['exit_code']
+      process_update(Proxy::Dynflow::Runner::Update.new(continuous_output, exit_code))
+    end
+
+    def kill_run
```
ezr-ondrej, line 49: I don't think this is ever being called, isn't this method called on the parent action now?

lib/smart_proxy_remote_execution_ssh.rb
```diff
@@ -2,6 +2,7 @@
 require 'smart_proxy_remote_execution_ssh/version'
 require 'smart_proxy_remote_execution_ssh/plugin'
 require 'smart_proxy_remote_execution_ssh/webrick_ext'
+require 'sequel'
```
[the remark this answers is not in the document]
adamruzicka: https://github.com/theforeman/smart_proxy_remote_execution_ssh/pull/64/files#diff-0dfc8e9074322ee9f207169162199e1dd15edd7005e183840e3b34c017cca364R6 I'll move it around

lib/smart_proxy_remote_execution_ssh/actions/pull_script.rb
adamruzicka: `::Proxy::Dynflow::Action::Runner` calls this when it receives `Cancel` event. As far as I know, Foreman currently doesn't know anything about the parent on the proxy side so it can't really be called there

adamruzicka:
Removed the cancellation for now, will open another PR once this goes in

PR #67, opened 2022-01-03
Fix cockpit loop method name

adamruzicka:
me, sometime in 2021: Oh, yeah, I'll definitely remember to eventually rename it *everywhere*


ares (approved):
Thanks @adamruzicka, merging!

PR #69, opened 2022-01-24
Fixes #34303 - Reap zombie ssh processes

adamruzicka:


ezr-ondrej (commented):
You say reap, but this actually gracefully waits for the ssh process to finish, that's what you mean ? :)
Or without wait the process becomes zombie? :thinking:

adamruzicka:
>  Or without wait the process becomes zombie?

Correct. It blocks until the child exits and then reaps it, same as `wait(2)`

ezr-ondrej (approved):
Thanks @adamruzicka :+1:

PR #70, opened 2022-02-01
Async SSH fixes

adamruzicka:
- Fix validation for mode
- Fixes #34368 - Make effective user work for async ssh


ezr-ondrej (approved):
I'll not pretend I know what's going on in the async-ssh, but this seems correct :+1:

PR #71, opened 2022-02-03
Fixes #34378 - Use process abstractions from sp-dynflow

adamruzicka:
Requires:
- https://github.com/theforeman/smart_proxy_dynflow/pull/103

TODO:
- [ ] bump sp-dynflow dependency once it gets released

### DIFF
--- a/lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
+++ b/lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
@@ -1,5 +1,6 @@
 require 'fileutils'
-require 'smart_proxy_dynflow/runner/command'
+require 'smart_proxy_dynflow/runner/process_manager_command'
+require 'smart_proxy_dynflow/process_manager'
 
 module Proxy::RemoteExecution::Ssh::Runners
   class EffectiveUserMethod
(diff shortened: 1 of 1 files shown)


ezr-ondrej:
This has a conflict now, but I believe this is the next one on the list? I'll take it for a spin, can you rebase? :)

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
```diff
@@ -387,9 +355,9 @@ def upload_file(local_path, remote_path)
     end
 
     def ensure_remote_directory(path)
-      exit_code, _output, err = run_sync("mkdir -p #{path}")
-      if exit_code != 0
-        raise "Unable to create directory on remote system #{path}: exit code: #{exit_code}\n #{err}"
+      pm = run_sync("mkdir -p #{path}")
+      if pm.status != 0
+        raise "Unable to create directory on remote system #{path}: exit code: #{pm.status}\n #{pm.stderr.to_s.chomp}"
```
ezr-ondrej, line 360: I've had the host down and this message appeared and I found it bit confusing, can we have something bit more telling?

To be frank next line was `ssh: connect to host gina-langhorn.oezr.local port 22: No route to host` so it is ok, it is just bit harder to read.

ezr-ondrej (commented):
Just one comment (just UX), in general I see it working as expected! :+1:

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
adamruzicka: I'm trying to address this in the followup pr https://github.com/theforeman/smart_proxy_remote_execution_ssh/pull/72

ezr-ondrej (approved):
Thanks @adamruzicka !

PR #72, opened 2022-02-03
Fixes #34363 - Improve error reporting

adamruzicka:
Includes:
- https://github.com/theforeman/smart_proxy_remote_execution_ssh/pull/71

Transitively requires:
- https://github.com/theforeman/smart_proxy_dynflow/pull/103

### DIFF
--- a/lib/smart_proxy_remote_execution_ssh/runners/polling_script_runner.rb
+++ b/lib/smart_proxy_remote_execution_ssh/runners/polling_script_runner.rb
@@ -131,7 +131,7 @@ module Proxy::RemoteExecution::Ssh::Runners
 
     def cleanup
       if @cleanup_working_dirs
-        assert_remote_command("rm -rf #{remote_command_dir}",
+        ensure_remote_command("rm -rf #{remote_command_dir}",
                               publish: true,
                               error: "Unable to remove working directory #{remote_command_dir} on remote system, exit code: %{exit_code}")
       end
--- a/lib/smart_proxy_remote_execution_ssh/runners/polling_script_runner.rb
+++ b/lib/smart_proxy_remote_execution_ssh/runners/polling_script_runner.rb
@@ -49,14 +49,13 @@ module Proxy::RemoteExecution::Ssh::Runners
     end
 
     def refresh
-      err = output = nil
       begin
-        _, output, err = run_sync("#{@user_method.cli_command_prefix} #{@retrieval_script}")
+        pm = run_sync("#{@user_method.cli_command_prefix} #{@retrieval_script}")
       rescue StandardError => e
         @logger.info("Error while connecting to the remote host on refresh: #{e.message}")
       end
 
-      process_retrieved_data(output, err)
+      process_retrieved_data(pm.stdout.to_s.chomp, pm.stderr.to_s.chomp)
     ensure
       destroy_session
     end
@@ -131,7 +130,11 @@ module Proxy::RemoteExecution::Ssh::Runners
     end
 
     def cleanup
-      run_sync("rm -rf #{remote_command_dir}") if @cleanup_working_dirs
+      if @cleanup_working_dirs
+        assert_remote_command("rm -rf #{remote_command_dir}",
+                              publish: true,
+                              error: "Unable to remove working directory #{remote_command_dir} on remote system, exit code: %{exit_code}")
+      end
     end
 
     def destroy_session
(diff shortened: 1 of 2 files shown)


adamruzicka:
The polling script runner relies on several tools being available on the remote machine (curl, gnu coreutils), maybe it would be a good idea to check their presence in preflight checks? :thinking:

lib/smart_proxy_remote_execution_ssh/runners/polling_script_runner.rb
```diff
@@ -126,7 +125,11 @@ def load_event_updates(event_data)
     end
 
     def cleanup
-      run_sync("rm -rf \"#{remote_command_dir}\"") if @cleanup_working_dirs
+      if @cleanup_working_dirs
+        assert_remote_command("rm -rf \"#{remote_command_dir}\"",
```
ezr-ondrej, line 129: what is `assert_remote_command`? assert should not run anything in my mind... it should just checks if something had run (had happend)...

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
```diff
@@ -380,9 +382,16 @@ def upload_file(local_path, remote_path)
...
-      exit_code, _output, err = run_sync("mkdir -p #{path}")
-      if exit_code != 0
-        raise "Unable to create directory on remote system #{path}: exit code: #{exit_code}\n #{err}"
+      assert_remote_command("mkdir -p #{path}",
+        publish: true,
+        error: "Unable to create directory #{path} on remote system, exit code: %{exit_code}"
+      )
+    end
+
+    def assert_remote_command(cmd, error: nil, **kwargs)
```
ezr-ondrej, line 391: Can we rename it to `ensure...` ?

ezr-ondrej (commented):
Thoughts: :exploding_head: 

I'll probably need to go from the dynflow PR to wrap my head around it, but if we could come up with little pieces of documentation while at it, it would be awesome (personally I believe we have quite a lot of code and we could get rid of some so I'm sad we are adding some, but it's just a feeling :)) )

lib/smart_proxy_remote_execution_ssh/runners/polling_script_runner.rb
adamruzicka: Right, it is "run the thing, check if it succeeded and raise an exception if not". If you thing `ensure_*` would depict it better then I'll change it

ezr-ondrej: I think so, it is not great, but IMO bit better then assure

ezr-ondrej, line 129: This is still assert, I'd expect it would fail, right?

adamruzicka: Correct

ezr-ondrej:
ok, lets resume here ^_^

adamruzicka:
rebased

lib/smart_proxy_remote_execution_ssh/runners/polling_script_runner.rb
adamruzicka: Oh, I misunderstood the comment previously, sorry about this

lib/smart_proxy_remote_execution_ssh/runners/polling_script_runner.rb
```diff
@@ -131,7 +130,11 @@ def load_event_updates(event_data)
     end
 
     def cleanup
-      run_sync("rm -rf #{remote_command_dir}") if @cleanup_working_dirs
+      if @cleanup_working_dirs
+        assert_remote_command("rm -rf #{remote_command_dir}",
```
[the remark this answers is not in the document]
adamruzicka: Right, now I got what you meant

ezr-ondrej (approved):
No, I do not and we can add more when we get some feedback on this, sorry I had the foolish idea to understand this completely :shrug:
Lets just go with "it works and looks good from what I see" :+1:

adamruzicka:
> I had the foolish idea to understand this completely 

May I help with this somehow?

ezr-ondrej:
> May I help with this somehow?

Too late, unless it breaks I do not want to understand (as I will not touch it unless I will need to) xD

PR #73, opened 2022-03-21
Fixes #34656 - Transfer working directory ownership to effective user

adamruzicka:
when async ssh is used


ezr-ondrej (approved):
LGTM, with only nitpicks and works as expected.
I was thinking if there might be some permission issues in some case, but I figured we shoul always be creating the base_dir, so it should not be issue never, right?

ezr-ondrej:
I've resolved my comments, but if those raise an eyebrow for someone else, we can follow-up :)

PR #74, opened 2022-03-22
Fixes #34595 - Control socket path length

pkoprda:
Socket path length is limited to 108 (or 104, on some platforms) characters, so we can't just use local working directory for that

### DIFF
--- a/lib/smart_proxy_remote_execution_ssh.rb
+++ b/lib/smart_proxy_remote_execution_ssh.rb
@@ -10,6 +10,7 @@ module Proxy::RemoteExecution
         validate_mode!
         validate_ssh_settings!
         validate_mqtt_settings!
+        validate_socket_path!
       end
 
       def private_key_file
@@ -94,6 +95,13 @@ module Proxy::RemoteExecution
 
       def requires_configured_ssh?
         %i[ssh ssh-async].include?(Plugin.settings.mode) || Plugin.settings.cockpit_integration
+      end
+
+      def validate_socket_path!
+        return unless Plugin.settings.mode == :'ssh' || Plugin.settings.mode == :'ssh-async'
+
+        socket_path = File.expand_path(Plugin.settings.socket_working_dir)
+        raise "Socket path #{socket_path} is too long" if socket_path.length > Plugin::SOCKET_PATH_MAX_LENGTH
       end
 
       def job_storage
--- a/lib/smart_proxy_remote_execution_ssh/plugin.rb
+++ b/lib/smart_proxy_remote_execution_ssh/plugin.rb
@@ -2,6 +2,10 @@ module Proxy::RemoteExecution::Ssh
   class Plugin < Proxy::Plugin
     SSH_LOG_LEVELS = %w[debug info error fatal].freeze
     MODES = %i[ssh async-ssh pull pull-mqtt].freeze
+    # Unix domain socket path length is limited to 104 (on some platforms) characters
+    # Socket path is composed of custom path (max 49 characters) + job id (37 characters)
+    # + offset(17 characters) + null terminator
+    SOCKET_PATH_MAX_LENGTH = 49
 
     http_rackup_path File.expand_path("http_config.ru", File.expand_path("../", __FILE__))
     https_rackup_path File.expand_path("http_config.ru", File.expand_path("../", __FILE__))
@@ -11,6 +15,7 @@ module Proxy::RemoteExecution::Ssh
                      :ssh_user                => 'root',
                      :remote_working_dir      => '/var/tmp',
                      :local_working_dir       => '/var/tmp',
+                     :socket_working_dir      => '/var/tmp',
                      :kerberos_auth           => false,
                      :cockpit_integration     => true,
                      # When set to nil, makes REX use the runner's default interval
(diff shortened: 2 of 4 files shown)


lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
```diff
@@ -340,6 +343,16 @@ def local_command_file(filename)
       File.join(ensure_local_directory(local_command_dir), filename)
     end
 
+    def socket_dir
+      File.join(@socket_working_dir, 'sockets', "foreman-ssh-cmd-#{@id}")
+    end
+
+    def socket_file
+      socket_path = File.join(ensure_local_directory(socket_dir), 'socket')
+      raise "Socket path #{socket_path} too long" if socket_path.length > SOCKET_PATH_MAX_LENGTH
```
adamruzicka, line 352: This happens at runtime, doesn't it? It would be better if we could check this at startup and disable the whole feature if we know it cannot work

pkoprda: Okay, thanks for suggestion :+1:  I guess it should be validated somewhere in this [file](https://github.com/theforeman/smart_proxy_remote_execution_ssh/blob/master/lib/smart_proxy_remote_execution_ssh.rb), right?

adamruzicka: Correct

lib/smart_proxy_remote_execution_ssh/plugin.rb
```diff
@@ -2,6 +2,7 @@ module Proxy::RemoteExecution::Ssh
   class Plugin < Proxy::Plugin
     SSH_LOG_LEVELS = %w[debug info error fatal].freeze
     MODES = %i[ssh async-ssh pull pull-mqtt].freeze
+    SOCKET_PATH_MAX_LENGTH = 18.freeze
```
adamruzicka, line 5: Could you add a comment here why it is 18?

adamruzicka, line 5: Also integers are `#frozen?` by default, no need to be explicit about it

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
```diff
@@ -340,6 +342,14 @@ def local_command_file(filename)
       File.join(ensure_local_directory(local_command_dir), filename)
     end
 
+    def socket_dir
+      File.join(@socket_working_dir, 'sockets', "foreman-ssh-cmd-#{@id}")
```
adamruzicka, line 346: The `sockets` and `foreman-ssh-cmd-$UUID` subdirectories are not really necessary. The user configures a directory in which the control sockets should live. I'd be ok with putting all the sockets directly into that directory with job id as the name for each of them. I don't think we'll need more than one control socket per job in the foreseeable future

lib/smart_proxy_remote_execution_ssh/plugin.rb
```diff
@@ -2,7 +2,10 @@ module Proxy::RemoteExecution::Ssh
   class Plugin < Proxy::Plugin
     SSH_LOG_LEVELS = %w[debug info error fatal].freeze
     MODES = %i[ssh async-ssh pull pull-mqtt].freeze
-    SOCKET_PATH_MAX_LENGTH = 18.freeze
+    # Unix domain socket path length is limited to 104 (on some platforms) characters
+    # Socket path is composed of custom path (max 49 characters) + job id (37 characters)
+    # + offset(17 characters) + null terminator
```
adamruzicka, line 7: What is the offset?

pkoprda: According to `man 7 unix` it should be the size of the address structure also known as `offsetof(struct sockaddr_un, sun_path)`, where `sun_path` is socket pathname. Maybe `sizeof sockaddr` would be better name?

lib/smart_proxy_remote_execution_ssh.rb
```diff
@@ -83,6 +84,11 @@ def validate_ssh_log_level!
         Plugin.settings.ssh_log_level = Plugin.settings.ssh_log_level.to_sym
       end
 
+      def validate_socket_path!
+        socket_path = Plugin.settings.socket_working_dir
+        raise "Socket path #{socket_path} too long" if socket_path.length > Plugin::SOCKET_PATH_MAX_LENGTH
```
adamruzicka, line 89:
```suggestion
        raise "Socket path #{socket_path} is too long" if socket_path.length > Plugin::SOCKET_PATH_MAX_LENGTH
```

lib/smart_proxy_remote_execution_ssh.rb
```diff
@@ -83,6 +84,11 @@ def validate_ssh_log_level!
         Plugin.settings.ssh_log_level = Plugin.settings.ssh_log_level.to_sym
       end
 
+      def validate_socket_path!
```
adamruzicka, line 87: This is only relevant if one of the ssh modes is used. Could we enforce this conditionally?

lib/smart_proxy_remote_execution_ssh.rb
adamruzicka, line 89: It should expand the path before checking the length. On my machine `~/aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa` resolves to `/home/aruzicka/aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa`. One passes this check, the other does not.

lib/smart_proxy_remote_execution_ssh/plugin.rb
adamruzicka: No, it is fine.

adamruzicka:
Needs a rebase

lib/smart_proxy_remote_execution_ssh.rb
adamruzicka: Ping

pkoprda:
Rebased and all :green_apple:

PR #75, opened 2022-03-23
Fixes #34667 - Add SSL support when connecting to mqtt broker

adamruzicka:
Use of SSL can be forced either way by explicitly setting `mqtt_tls` setting. If unset, it gets used if certificate, private key and CA
certificate are available. Currently it reuses the `ssl_*` set of certs smart proxy has.

### DIFF
--- a/lib/smart_proxy_remote_execution_ssh/actions/pull_script.rb
+++ b/lib/smart_proxy_remote_execution_ssh/actions/pull_script.rb
@@ -90,9 +90,9 @@ module Proxy::RemoteExecution::Ssh::Actions
     def with_mqtt_client(&block)
       MQTT::Client.connect(settings.mqtt_broker, settings.mqtt_port,
                            :ssl => settings.mqtt_tls,
-                           :cert_file => ::Proxy::SETTINGS.ssl_certificate,
-                           :key_file => ::Proxy::SETTINGS.ssl_private_key,
-                           :ca_file => ::Proxy::SETTINGS.ssl_ca_file,
+                           :cert_file => ::Proxy::SETTINGS.foreman_ssl_cert,
+                           :key_file => ::Proxy::SETTINGS.foreman_ssl_key,
+                           :ca_file => ::Proxy::SETTINGS.foreman_ssl_ca,
                            &block)
     end
 
--- a/lib/smart_proxy_remote_execution_ssh/actions/pull_script.rb
+++ b/lib/smart_proxy_remote_execution_ssh/actions/pull_script.rb
@@ -82,9 +82,18 @@ module Proxy::RemoteExecution::Ssh::Actions
     end
 
     def mqtt_notify(payload)
-      MQTT::Client.connect(settings.mqtt_broker, settings.mqtt_port) do |c|
+      with_mqtt_client do |c|
         c.publish(mqtt_topic, JSON.dump(payload), false, 1)
       end
+    end
+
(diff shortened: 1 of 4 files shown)


ezr-ondrej:
Is it worth adding the setting to the example config?

ezr-ondrej:
@ehelms care to weigh in? My CA knowledge on Foreman infra is about zero so can't judge this too much.

lib/smart_proxy_remote_execution_ssh/actions/pull_script.rb
```diff
@@ -82,11 +82,20 @@ def mqtt_start(otp_password)
...
         c.publish(mqtt_topic, JSON.dump(payload), false, 1)
       end
     end
 
+    def with_mqtt_client(&block)
+      MQTT::Client.connect(settings.mqtt_broker, settings.mqtt_port,
+                           :ssl => settings.mqtt_tls,
+                           :cert_file => ::Proxy::SETTINGS.ssl_certificate,
+                           :key_file => ::Proxy::SETTINGS.ssl_private_key,
+                           :ca_file => ::Proxy::SETTINGS.ssl_ca_file,
```
ehelms, line 95: I have been thinking about certificates and this raises a good question: should we use the smart-proxy server certificates or the foreman client certificates (`:foreman_ssl_ca`, `:foreman_ssl_cert`, `:foreman_ssl_key`) to talk to mosquitto?

A few things to consider with this.

1) In a standard Foreman installation, this should be all Puppet certificates so it doesn't really matter.
2) In a standard Katello installation, this should be all certificates from the default CA so it doesn't really matter.

And then there is the dragon scenarios:

3) If the user provides custom server certificates in the Katello scenario:
   a) the smart-proxy server certificates will be the custom certificates and mosquitto will need to be configured with the CA of those certificates to accept them. 
   b) the smart-proxy Foreman client certificates will be from the default CA and if used, mosquitto would need to be configured with the default CA

That leads in to what certificates should we configure mosquitto with: ones derived only from the default CA or the same server certificates running the smart-proxy?

Foreman has a similar scenario if the user is using Puppet certificates but has opted to provide the smart-proxy with server certificates from a different CA.

@ekohl @evgeni your thoughts are appreciated, apologies for any headaches this may cause

evgeni: This is one of the headaches I already had (thanks to a very special tree).
Either Mosquitto or Go's TLS stack was very picky about which certificate was used where, based on flags present. So using the non-client cert didn't work and "solved" the selection for me.

adamruzicka: Switched to client certs

adamruzicka:
Anything else to do here?

settings.d/remote_execution_ssh.yml.example
```diff
@@ -24,3 +24,8 @@
 # MQTT configuration, need to be set if mode is set to pull-mqtt
 # :mqtt_broker: localhost
 # :mqtt_port: 1883
+
+# Use of SSL can be forced either way by explicitly setting mqtt_tls setting. If
+# unset, SSL gets used if smart-proxy's ssl_certificate, ssl_private_key and
+# ssl_ca_file settings are set available.
```
ehelms, line 30: Does this also need updating if using client certificates?

adamruzicka:
Thank you everyone!

PR #81, opened 2022-04-28
Fixes #34833 - Extend the accepted job update format

adamruzicka:
A `'1'` version update is

```
{ "version": "1", "updates": [OutputUpdate | ExitUpdate] }
```

An `OutputUpdate` is:
```
{ "type":      "output",
  "timestamp": "$TIMESTAMP",
  "stream":    "$STREAM",
  "content":   "$CONTENT" }
```

where
`$TIMESTAMP` is ISO8601 timestamp of when the output was emitted
`$STREAM` provides supplemental information about from where the output was
emitted, usually this would be either `"stdout"` or `"stderr"`
`$CONTENT` is the emitted output itself

An `ExitUpdate` is:
```
{ "type": "exit", "exit_code": $EXIT_CODE }
```

where
`$EXIT_CODE` is the exit code of the script being run. Once a `'1'` version
update containing an `ExitUpdate` is processed, the job is marked as done.


ofedoren (approved):
Thanks, @adamruzicka, LGTM.

PR #82, opened 2022-05-05
Be more explicit about dealing with connections

adamruzicka:
This allows us to:
- try using both password and key passphrase
- have clearer idea what's going on
- have better logging
- decouple connection handling logic from what we want to run on the remote end

### DIFF
--- a/lib/smart_proxy_remote_execution_ssh/multiplexed_ssh_connection.rb
+++ b/lib/smart_proxy_remote_execution_ssh/multiplexed_ssh_connection.rb
@@ -1,0 +1,185 @@
+require 'smart_proxy_remote_execution_ssh/command_logging'
+
+module Proxy::RemoteExecution::Ssh::Runners
+  class SensitiveString
+    def initialize(value, mask: '*****')
+      @value = value
+      @mask = mask
+    end
+
+    def inspect
+      '"' + to_s + '"'
+    end
+
+    def to_s
+      @mask
+    end
+
+    def to_str
+      @value
+    end
+  end
+
+  class AuthenticationMethod
+    attr_reader :name
+    def initialize(name, prompt: nil, password: nil)
+      @name = name
+      @prompt = prompt
+      @password = password
+    end
+
+    def ssh_command_prefix
+      return [] unless @password
+
+      prompt = ['-P', @prompt] if @prompt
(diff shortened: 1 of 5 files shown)


adamruzicka:
@jhoblitt I'm reworking how we handle and reuse ssh connections. While at it, I also added a bit more logging, would something like this work for you https://gist.github.com/adamruzicka/866a82bb12c1f89da65db218c0cef4b8 or is it still not detailed enough?

jhoblitt:
@adamruzicka That looks great.  What I really appreciate is being able to cut'n'paste command lines verbatim.  Far more than once, I've introduced typos trying to reconstruct command args while debugging.

adamruzicka:
@ofedoren Could you please take a look at this?

lib/smart_proxy_remote_execution_ssh/multiplexed_ssh_connection.rb
```diff
@@ -0,0 +1,195 @@
...
+      @name = name
+      @prompt = prompt
+      @password = password
+    end
+
+    def ssh_command_prefix
+      return [] unless @password
+
+      prompt = ['-P', @prompt] if @prompt
+      [{'SSHPASS' => SensitiveString.new(@password)}, '/usr/bin/sshpass', '-e', prompt].compact
```
ofedoren, line 35: How is it ensured that `to_str` will be called on `SensitiveString` later instead of `to_s`?..

ofedoren (approved):
Thanks, @adamruzicka. Codewise LGTM and seems to be working. I don't think I have any concerns. Just one question inline.

lib/smart_proxy_remote_execution_ssh/multiplexed_ssh_connection.rb
adamruzicka: Good question. When a hash is being printed it seems to be calling `#inspect` on the values and when the environment is being processed by `spawn` it seems to use `to_str` on them. Since `spawn` is implemented directly in C I did not dare to follow what it does and how it actually works under the hood/ It may prove insufficient if we later add more features, but for now it seems to work reasonably well

ofedoren: ACK from me then :)

PR #83, opened 2022-05-05
Fixes #34869 - Suppress messages about shared connection being closed

adamruzicka:
a workaround until https://github.com/theforeman/smart_proxy_remote_execution_ssh/pull/82 is done


adamruzicka:
Looking at it once more, it seems the patch should not be needed for 0.5.3 at all? We seem to be actually ignoring stderr from the main command, which is probably not correct, but avoids the issue.

PR #88, opened 2022-07-15
Cockpit fixes

adamruzicka:


adamruzicka:
That's a fair question. Originally (before we dropped net/ssh), we were using net/ssh directly when dealing with cockpit bits. When we moved from `net/ssh` to using system ssh, I used `ScriptRunner` just because I needed some parts of its functionality (connection handling), which weren't exposed anywhere else. Now that we have `MultiplexedSSHConnection`, we can use it directly and It even fits the use case better. One could even say moving connection handling bits to its own class and then reusing it here is a long overdue refactoring that we needed.

ofedoren (approved):
Thanks. In that case let's get this in :)

PR #89, opened 2022-07-19
Fixes #35245 - Override ProxyCommand to connect directly

adamruzicka:


Ron-Lavi (approved):
Looks safe to me :)

adamruzicka:
For reference, we do the same thing for ansible https://github.com/theforeman/puppet-foreman_proxy/commit/2e11f6810d8b4e7e8ce28f91bdf6a534c0f3e2f4

Ron-Lavi:
Cool, I'd say let's merge it... though I don't have permission.
Maybe @ofedoren can take a look?

ofedoren (approved):
Thanks, @adamruzicka, @Ron-Lavi! Seems I have permissions all of a sudden.

BenArgyle:
This breaks things for me.  I use a ProxyJump directive in ~foreman-proxy/.ssh/config to get through a NAT to all the hosts inside, and now that a/any Proxy* directive is hardcoded, my ProxyJump directive is ignored.

adamruzicka:
> This breaks things for me

I'm sorry to hear that.

What would you suggest? We could make it configurable the same way ansible does, but it opens door to people passing in arguments that could break things

BenArgyle:
I wish I was smart enough to suggest an agreeable-to-all fix.  I'm engaging with Red Hat Support (I'm a Red Hat Satellite 6.12.2 user, rather than Foreman) about this issue at the moment.  Part of me thinks making it configurable would be a good idea; there are tens of thousands of other things in the open (and closed!) source world that work on the equivalent of "it won't work if you do something stupid".  Anything that has a configuration file that's user-editable runs the same risk.

But again, I'm not a programmer/developer/person who understands all the complexities/background of this.  I just want it to work (like it used to) (-:

snagoor:
Hello @adamruzicka,

Hardcoding ProxyCommand  wouldn't help support all possible scenarios, for example jump host configurations configured in `~foreman-proxy/.ssh/config` will be ignored. 

The change that was intended in this PR can easily be fixed in the documentation guide to instruct end-users while configuring IDM auth use the `--no-ssh` switch with `ipa-client-install` command. Using this option `ipa-client-install` will not configure a ProxyCommand directive in the global SSH config.

IMHO, this patch should be reverted to support all scenarios and a new documentation bug to be opened for https://projects.theforeman.org/issues/35245

gvde:
I have opened a new bug https://projects.theforeman.org/issues/36456 for the lack of proxy host support.

Using `--no-ssh` isn't really an option either, because it will prevent known hosts taken from ipa...

PR #90, opened 2022-07-20
Rate limit

adamruzicka:
Job storage now has an additional column indicating whether the job is
running on a client. If a limit is set and reached and a client requests
another job, it will get 429 - Too many requests.

It would be cleaner to set some HTTP headers, but as far as I know
yggdrasil does not care about them.

Includes:
- https://github.com/theforeman/smart_proxy_remote_execution_ssh/pull/87


adamruzicka:
Closing in favor of #98

PR #92, opened 2022-08-22
Fixes #35406 - Honor time to pickup when pull mode is used

adamruzicka:


ofedoren (approved):
Thanks, @adamruzicka, now I can see the time out error: 
ScreenShot-1662642595936 And the job is no longer pending.

Let's get this in.

PR #94, opened 2022-09-20
Fixes #35541 - Mark session as closed after run_sync when using async_ssh

adamruzicka:


ianballou (approved):
Tested against a 6.11 machine, works well.  After applying, I was able to run async ssh commands on my el8 host.

adamruzicka:
Thank you @ianballou !

PR #95, opened 2022-09-20
Fixes #35542 - Fix async ssh

adamruzicka:
> lib/.../script_runner.rb

Without this change, the job would hang around forever, as the output of the command we run to check the job on the remote end would be discarded.

> lib/.../polling_script_runner.rb

This kwarg does not exist anymore.

### DIFF
--- a/lib/smart_proxy_remote_execution_ssh/runners/polling_script_runner.rb
+++ b/lib/smart_proxy_remote_execution_ssh/runners/polling_script_runner.rb
@@ -133,7 +133,6 @@ module Proxy::RemoteExecution::Ssh::Runners
     def cleanup
       if @cleanup_working_dirs
         ensure_remote_command("rm -rf #{remote_command_dir}",
-                              publish: true,
                               error: "Unable to remove working directory #{remote_command_dir} on remote system, exit code: %{exit_code}")
       end
     end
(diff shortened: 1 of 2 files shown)


ianballou (approved):
Tested an ssh-async job and it ran fine with these changes. I created a file on the host successfully. I did see an error from the proxy, but I think it's unrelated:

```
2022-09-23T19:39:52  [E] <OpenSSL::SSL::SSLError> SSL_accept returned=1 errno=0 state=error: tlsv1 alert unknown ca
	/usr/share/ruby/webrick/server.rb:299:in `accept'
	/usr/share/ruby/webrick/server.rb:299:in `block (2 levels) in start_thread'
	/usr/share/ruby/webrick/utils.rb:263:in `timeout'
	/usr/share/ruby/webrick/server.rb:297:in `block in start_thread'
	/usr/share/gems/gems/logging-2.3.0/lib/logging/diagnostic_context.rb:474:in `block in create_with_logging_context'
```

adamruzicka:
> I did see an error from the proxy, but I think it's unrelated:

That looks like some kind of misconfiguration, certs are hard to get right.

adamruzicka:
Thank you @ianballou !

PR #101, opened 2023-03-10
Upgrade to Ruby 3.1

archanaserver:
Adding Ruby v3.0 & v3.1 support.


lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
```diff
@@ -180,7 +179,8 @@ def prepare_start
       @output_path = File.join(File.dirname(@remote_script), 'output')
       @exit_code_path = File.join(File.dirname(@remote_script), 'exit_code')
       @pid_path = File.join(File.dirname(@remote_script), 'pid')
-      @remote_script_wrapper = upload_data("echo $$ > #{@pid_path}; exec \"$@\";", File.join(File.dirname(@remote_script), 'script-wrapper'), 555)
+      @remote_script_wrapper = upload_data("echo $$ > #{@pid_path}; exec \"$@\";",
+  File.join(File.dirname(@remote_script), 'script-wrapper'), 555)
```
adamruzicka, line 183: This looks wrong

archanaserver: @adamruzicka i can't get you point here, can you expand a little more your views?

adamruzicka: Well, rubocop may like it, but if suddenly indentation goes from 6 spaces to 2 in the middle of a method call, I'd say that's wrong.

I'd expect either
```ruby
      @remote_script_wrapper = upload_data("echo $$ > #{@pid_path}; exec \"$@\";",
                                           File.join(File.dirname(@remote_script), 'script-wrapper'), 555)
```

or
```ruby
  @remote_script_wrapper = upload_data(
    "echo $$ > #{@pid_path}; exec \"$@\";",
    File.join(File.dirname(@remote_script), 'script-wrapper'),
    555)
```

Having the second line indented way less than the first one, even though they form a single logical unit feels wrong.

archanaserver: @adamruzicka Yes, this make sense now. Updating the changes. Thanks!

adamruzicka (approved):
LGTM

adamruzicka:
Thank you @archanaserver !

PR #102, opened 2023-03-10
Fixes #37593 - Prune dead code

adamruzicka:


lib/smart_proxy_remote_execution_ssh/actions/pull_script.rb
```diff
@@ -47,10 +47,6 @@ def run(event = nil)
     end
 
     def init_run
-      otp_password = if with_mqtt?
```
ofedoren: Fore context: why we're dropping OTP password? And why we're leaving it here: https://github.com/theforeman/smart_proxy_remote_execution_ssh/blob/master/lib/smart_proxy_remote_execution_ssh/runners/polling_script_runner.rb#L27?

adamruzicka: We're dropping it only from pull mode, since it *requires* to use certs anyway. We're keeping it for polling ssh where we push the job to the client, disconnect and expect the client to be able to call back without any certs.

adamruzicka:
> Also, does it deserve a RM ticket with some description regarding why this is dead code?

I guess it does. I originally wanted to do something much smaller, but it kind of snowballed

ofedoren (approved):
Thanks a lot, @adamruzicka !

PR #103, opened 2023-07-26
Fixes #36615 - Resume dispatcher actor on exception

adamruzicka:
The previous implementation wrongly assumed that the :resume! part of behaviour applied to the actor while in fact it applies to child actors supervised by the actor with this behaviour.

This change spawns the dispatcher actor under a supervisor, which resumes the dispatcher in case the dispatcher raises an exception.

### DIFF
--- a/lib/smart_proxy_remote_execution_ssh/mqtt/dispatcher.rb
+++ b/lib/smart_proxy_remote_execution_ssh/mqtt/dispatcher.rb
@@ -2,15 +2,39 @@ require 'concurrent'
 require 'mqtt'
 
 class Proxy::RemoteExecution::Ssh::MQTT
+  class DispatcherSupervisor < Concurrent::Actor::RestartingContext
+    def initialize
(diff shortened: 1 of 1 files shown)


ofedoren (approved):
Thanks, @adamruzicka, code-wise looks good, no questions. Although, I'm not quite sure how could I do the "reproduce->test fixed" loop.

lib/smart_proxy_remote_execution_ssh/mqtt/dispatcher.rb
```diff
@@ -2,15 +2,39 @@
 require 'mqtt'
 
 class Proxy::RemoteExecution::Ssh::MQTT
+  class DispatcherSupervisor < Concurrent::Actor::RestartingContext
+    def initialize()
```
ofedoren, line 6: That was a long time I've seen a Ruby method definition with empty param list using brackets :)

PR #106, opened 2023-08-07
Fixes #36643 - Use Minitest namespace

evgeni:


ekohl (changes requested):
I already submitted https://github.com/theforeman/smart_proxy_remote_execution_ssh/pull/105. I've just updated the commit message to use the Redmine issue number.

evgeni:
indeed, yours is better

PR #107, opened 2023-08-07
Fixes #36647 - Fix su{,do} and preflight checks

adamruzicka:
If privilege escalation is scoped to only the necessary files, the effective user preflight check might fail. This change reuses one of the files used later in the execution to perform the preflight checks and not require additional commands to be allowed.


ofedoren (approved):
Thanks, @adamruzicka, AFAICT, LGTM :)

PR #108, opened 2023-08-08
Fixes #36655 - Fail the job if exit code is not written

adamruzicka:
In case the disk gets full, the exit code file might not be there at all or might be empty.


ofedoren (approved):
Thanks, @adamruzicka, let's get this in.

PR #109, opened 2023-08-09
Fixes #36662 - Check host keys

adamruzicka:


nofaralfasi:
It works well for me. I tested it on a remote host that is registered in my environment, and I have previously executed jobs on it. However, after changing this remote host's public key, I encountered connectivity issues when attempting to run a remote job on it.

lib/smart_proxy_remote_execution_ssh/multiplexed_ssh_connection.rb
```diff
@@ -119,10 +127,10 @@ def try_auth_method(method)
       if pm.status.zero?
         logger.debug("Established connection using authentication method #{method.name}")
         @socket = socket_file
-        true
+        [true, pm]
```
ofedoren, line 130: I know, we all rather write in Go :D

But looking at this method, it looks weird (from Ruby perspective) in both cases: old and new.

What if we make `try_auth_method` always return `pm`? `true/false` seems to not be helpful anyway. Especially when the name of the method is without `?`.

lib/smart_proxy_remote_execution_ssh/multiplexed_ssh_connection.rb
```diff
@@ -67,12 +69,18 @@ def initialize(options, logger:)
     def establish!
       @available_auth_methods ||= available_authentication_methods
       method = @available_auth_methods.find do |method|
-        if try_auth_method(method)
+        result, pm = try_auth_method(method)
+        method.errors = pm.stderr
+        if result
           @available_auth_methods.unshift(method).uniq!
           true
         end
```
ofedoren, line 77: With the changes to `try_auth_method` this could be:

```suggestion
        result = try_auth_method(method)
        method.errors = result.stderr
        if result.status.zero?
          @available_auth_methods.unshift(method).uniq!
          true
        end
```

ofedoren (commented):
Thanks, @adamruzicka, few codewise suggestions inline.

pondrejk (approved):
ack, checked on patched stream

ofedoren (approved):
Thanks, all, merging :)

PR #112, opened 2024-02-19
packit updates

adamruzicka:
- use yum.tfm.org
- use https
- use rhel-8 instead of centos-stream-8
- build for el9


ekohl (approved):

adamruzicka:
Thank you @ekohl!

PR #113, opened 2024-05-16
Fixes #37475 - Avoid kwargs in action methods

adamruzicka:
It used to work on Ruby 2, but broke with Ruby 3.


pondrejk (approved):
checked on patched stream 59 capsule on rhel9, I confirm the original issue got fixed by this PR

adamruzicka:
Thank you @pondrejk !

PR #114, opened 2024-05-16
Run tests and rubocop in GHA

adamruzicka:


ofedoren (approved):

adamruzicka:

.github/workflows/test.yml
```diff
@@ -0,0 +1,18 @@
...
+  cancel-in-progress: true
+
+jobs:
+  rubocop:
+    name: Rubocop
+    uses: theforeman/actions/.github/workflows/rubocop.yml@v0
+
+  test:
+    name: Tests
+    uses: theforeman/actions/.github/workflows/test-gem.yml@v0
```
ekohl, line 16: We have a dedicated Smart Proxy test workflow. Would you mind updating to that? Biggest difference is that we can enforce certain Ruby version support and fail checks if that isn't there.

PR #115, opened 2024-05-23
Cancellation improvements

adamruzicka:


lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
```diff
@@ -160,6 +164,16 @@ def trigger(*args)
       run_async(*args)
     end
 
+    def detect_capabilities
```
ofedoren, line 167: It implies multiple different capabilities, but this method only checks for `unshare` support :/

ofedoren (approved):
Thanks, @adamruzicka, LGTM, just a naming concern :/

pondrejk (approved):
checked on patched stream, fixes the original issue

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
adamruzicka: Eventually there might be more, but I can rename it if you feel strongly about it

ofedoren (approved):
Thanks, @adamruzicka, I have nothing to add and thanks @pondrejk for testing out.

PR #116, opened 2024-06-18
Fixes #37570 - Fix su with password

adamruzicka:
Apparently, due to another shell in place, the password didn't reach su even though we were writing it in.

### DIFF
--- a/lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
+++ b/lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
@@ -118,7 +118,6 @@ module Proxy::RemoteExecution::Ssh::Runners
       @first_execution = options.fetch(:first_execution, false)
       @user_method = user_method
       @options = options
-      @supports_unshare = false
     end
 
     def self.build(options, suspended_action:)
@@ -150,7 +149,6 @@ module Proxy::RemoteExecution::Ssh::Runners
       @connection = MultiplexedSSHConnection.new(@options.merge(:id => @id), logger: logger)
       @connection.establish!
       preflight_checks
-      detect_capabilities
       prepare_start
       script = initialization_script
       logger.debug("executing script:\n#{indent_multiline(script)}")
@@ -162,16 +160,6 @@ module Proxy::RemoteExecution::Ssh::Runners
 
     def trigger(*args)
       run_async(*args)
-    end
-
-    def detect_capabilities
-      script = cp_script_to_remote("#!/bin/sh\nexec #{UNSHARE_PREFIX} true")
-      begin
-        ensure_remote_command(script)
-        @supports_unshare = true
-      rescue; end
-      # The path should already be escaped
-      ensure_remote_command("rm #{script}")
     end
 
     def preflight_checks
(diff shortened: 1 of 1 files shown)


adamruzicka:
Hold on, this seems to break cancellation

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
```diff
@@ -195,27 +195,28 @@ def prepare_start
...
+      su_method = @user_method.instance_of?(SuUserMethod)
+      wrapper = <<~SCRIPT
+        if [ "$1" == "inner" ]; then
+          echo \$$ > #{@pid_path}
+          (
+            #{@user_method.cli_command_prefix}#{su_method ? "'exec #{@remote_script} < /dev/null '" : "#{@remote_script} < /dev/null"}
+            echo \$? >#{@exit_code_path}
+          ) | tee #{@output_path}
+        else
+          exec #{UNSHARE_PREFIX} "$0" inner
```
adamruzicka, line 207: this will need something like
```suggestion
          UNSHARE=''
          if #{UNSHARE_PREFIX} true >/dev/null 2>/dev/null; then
              UNSHARE='#{UNSHARE_PREFIX}'
          fi
          exec $UNSHARE "$0" inner
```

adamruzicka (commented):
kudos to @pondrejk for pointing this out

pondrejk (approved):
Ack from the functional side, running a job using su method with a password succedes on stream build with this patch

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
```diff
@@ -164,16 +163,6 @@ def trigger(*args)
       run_async(*args)
     end
 
-    def detect_capabilities
```
ofedoren, line 167: Doesn't it revert https://github.com/theforeman/smart_proxy_remote_execution_ssh/pull/115/files ?

ofedoren (commented):
Thanks, @adamruzicka and @pondrejk, just a question to be sure:

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
adamruzicka: Not really, it just got moved around a little bit. The unshare detection now happens here https://github.com/theforeman/smart_proxy_remote_execution_ssh/pull/116/files#diff-f7a31028d1cbfee3429cdcb724c2fd35effcf3b665e39aec9c348ee015415313R196-R199

adamruzicka: done

ofedoren (approved):
Thanks, @adamruzicka and @pondrejk !

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
```diff
@@ -195,27 +183,32 @@ def prepare_start
...
+      wrapper = <<~SCRIPT
+        if [ "$1" == "inner" ]; then
+          echo \$$ > #{@pid_path}
+          (
+            #{@user_method.cli_command_prefix}#{su_method ? "'exec #{@remote_script} < /dev/null '" : "#{@remote_script} < /dev/null"}
+            echo \$? >#{@exit_code_path}
+          ) | tee #{@output_path}
+        else
+          UNSHARE=''
+          if #{UNSHARE_PREFIX} true >/dev/null 2>/dev/null; then
```
lhellebr, line 196: Does this just test if there is unshare installed and if there isn't, we're happy with just not using it?

adamruzicka: Correct

PR #117, opened 2024-08-27
Fixes #36456 - Use user-defined ProxyCommand if there is one

adamlazik1:


gvde:
I think, if you want to go down that road and check the configuration, you would have to run to actual ssh command with  `ssh -G` to see the resulting configuration. That includes the global configuration as well as Match blocks. Then you could be sure if ProxyCommand has been set or not.

Otherwise, I would still vote for adding the ProxyCommand option into ~foreman-proxy/.ssh/config. Then users can add Match blocks where needed.

adamlazik1:
Thank you for the tip, but I probably cannot check with `ssh -G`, since older versions of `ipa-client-install` were putting proxycommand into `/etc/ssh/ssh_config`, which subsequently broke things here. That is the reason we have `ProxyCommand=none` hardcoded as of now. However, I realize that the solution I provided is far from ideal as well. I am looking more into putting it into proxy ssh config file and what that would do to existing proxy instances. I will have more info on that later this week.

gvde:
> Thank you for the tip, but I probably cannot check with `ssh -G`, since older versions of `ipa-client-install` were putting proxycommand into `/etc/ssh/ssh_config`, which subsequently broke things here.

I don't quite understand. `ssh -G` will tell you the exact `ProxyCommand` configuration used for a specific ssh command. Thus if you pass the same arguments to `ssh -G` as you pass to `ssh` later it'll tell you exactly the active configuration. Regardless of where it is defined.

`ipa-client-install` still adds a ProxyCommand to the ssh config, even though it's in `/etc/ssh/ssh_config.d/04-ipa.conf` and the `ProxyCommand` is protected by an `Match exec true` which prevents execution if the executing ssh account is nologin.

What is the exact problem why you want to disable ProxyCommand to begin with and why can't you let the user decide if or how to implement this?

Putting it into the ssh config is definitively better. As it is running as foreman-proxy you could always put a `Match localhost foreman-proxy` in the right position to change whatever foreman-proxy is using... That should reduce interference with other configuration set for the system.

adamlazik1:
> I don't quite understand. ssh -G will tell you the exact ProxyCommand configuration used for a specific ssh command. Thus if you pass the same arguments to ssh -G as you pass to ssh later it'll tell you exactly the active configuration. Regardless of where it is defined.

Yes, however; this could be a problem. If someone has foreman instance that is configured with ipa server and it was configured at a time when `ipa-client-install` put ProxyCommand in /etc/ssh/ssh_config, checking config with `ssh -G` would reveal that there is proxycommand set and that it shouldn't be overridden, even though this very ProxyCommand is the reason we had to hardcode `ProxyCommand=none` in the first place. Subsequently, ansible jobs could start breaking on some systems ([BZ#1650103](https://bugzilla.redhat.com/show_bug.cgi?id=1650103)). 

With all that being said, we had a discussion with @adamruzicka and decided that since this is the issue of the old ipa tooling and not an issue of foreman, we should just remove the hardcoded proxycommand, and then explain to users who might have the proxycommand from the old ipa-client-install (and at the same time do not have `/sbin/nologin` - this condition is required for the bug to occur) that they should just remove this line from /etc/ssh/ssh_config.

PR #118, opened 2024-09-04
Refs #37570 - use correct POSIX shell syntax for the wrapper script

evgeni:
Comparison with `==` is a bashism, POSIX [/test wants a single `=`.
The code worked on EL as there /bin/sh is bash, but Debian and friends
have dash which is POSIX but does not support `[ "a" == "b" ]`.

### DIFF
--- a/lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
+++ b/lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
@@ -185,7 +185,7 @@ module Proxy::RemoteExecution::Ssh::Runners
       @pid_path = File.join(File.dirname(@remote_script), 'pid')
       su_method = @user_method.instance_of?(SuUserMethod)
       wrapper = <<~SCRIPT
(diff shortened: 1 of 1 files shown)


lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
```diff
@@ -185,7 +185,7 @@ def prepare_start
       @pid_path = File.join(File.dirname(@remote_script), 'pid')
       su_method = @user_method.instance_of?(SuUserMethod)
       wrapper = <<~SCRIPT
-        if [ "$1" == "inner" ]; then
+        if [ "$1" = "inner" ]; then
```
evgeni, line 188: With `==` you get
```
[: inner: unexpected operator
```
and the `if` fails, executing the `else` part (with the `exec`) again and again and again in a loop.

adamruzicka (approved):

adamruzicka:
Thank you @evgeni !

PR #122, opened 2024-12-13
Use non-deprecated license identifier

adamruzicka:


adamruzicka:
Ah, I already did this in #120 , closing.

PR #126, opened 2025-06-05
Refs #38478 - Introduce SSH CA certificate support

adamlazik1:


lib/smart_proxy_remote_execution_ssh/multiplexed_ssh_connection.rb
```diff
@@ -59,7 +59,11 @@ def initialize(options, logger:)
       @key_passphrase = options.fetch(:secrets, {}).fetch(:key_passphrase, nil)
       @host_public_key = options.fetch(:host_public_key, nil)
       @verify_host = options.fetch(:verify_host, nil)
+      @use_ssh_certificates = options.fetch(:use_ssh_certificates, false)
```
adamruzicka, line 62: I'd drop this

lib/smart_proxy_remote_execution_ssh/plugin.rb
```diff
@@ -12,6 +12,7 @@ class Plugin < Proxy::Plugin
 
     settings_file "remote_execution_ssh.yml"
     default_settings :ssh_identity_key_file   => '~/.ssh/id_rsa_foreman_proxy',
+                     :ssh_ca_known_hosts_file => '~/.ssh/ca_known_hosts',
```
adamruzicka, line 15: Usually we also add all the newly introduced settings which don't have a default value as comments

lib/smart_proxy_remote_execution_ssh/multiplexed_ssh_connection.rb
```diff
@@ -154,9 +158,11 @@ def establish_ssh_options
       ssh_options << "-o User=#{@ssh_user}"
       ssh_options << "-o Port=#{@ssh_port}" if @ssh_port
       ssh_options << "-o IdentityFile=#{@client_private_key_file}" if @client_private_key_file
+      ssh_options << "-o CertificateFile=#{@client_ca_cert_file}" if @use_ssh_certificates && @client_ca_cert_file
       ssh_options << "-o IdentitiesOnly=yes"
       ssh_options << "-o StrictHostKeyChecking=accept-new"
-      ssh_options << "-o UserKnownHostsFile=#{prepare_known_hosts}" if @host_public_key
+      ssh_options << "-o UserKnownHostsFile=#{@client_ca_known_hosts_file}" if @use_ssh_certificates && @client_ca_known_hosts_file
+      ssh_options << "-o UserKnownHostsFile=#{prepare_known_hosts}" if !@use_ssh_certificates && @host_public_key
```
adamruzicka, line 165: This will need to be handled a little bit differently.

I'd say that if Foreman knows the host's public key (which I'm afraid never happens), it shouldn't rely on anything else.

adamruzicka, line 165: If `client_ca_cert_file` are `client_ca_known_hosts_file` are set, but the client doesn't accept them, will it fallback to traditional non-ca pubkey authentication?

lib/smart_proxy_remote_execution_ssh/multiplexed_ssh_connection.rb
adamlazik1: Dropped.

lib/smart_proxy_remote_execution_ssh/plugin.rb
adamlazik1: Added.

lib/smart_proxy_remote_execution_ssh/multiplexed_ssh_connection.rb
adamlazik1: I will need to test this once I have the ssh cert setup in a functional state.

adamlazik1: Updated.

adamlazik1: It appears that the known hosts file does not need to exist at all, but the authentication seems to fail if the client does not accept the certificate. I will do some further testing, but I would say that authentication attempt failing upon not accepting the certificate is the desired outcome in regards to security.

lib/smart_proxy_remote_execution_ssh.rb
```diff
@@ -21,6 +21,15 @@ def public_key_file
         File.expand_path("#{private_key_file}.pub")
       end
 
+      def cert_file
+        File.expand_path("#{private_key_file}-cert.pub")
+      end
```
adamlazik1, line 26: Should we require this strict naming convention for cert file
similarly as we do for public key file, or should we make this an editable
value in settings, with this format being the possible default?

lib/smart_proxy_remote_execution_ssh/multiplexed_ssh_connection.rb
```diff
@@ -60,6 +60,10 @@ def initialize(options, logger:)
       @host_public_key = options.fetch(:host_public_key, nil)
       @verify_host = options.fetch(:verify_host, nil)
       @client_private_key_file = settings.ssh_identity_key_file
+      @client_ca_known_hosts_file = settings.ssh_ca_known_hosts_file
+      @client_ca_public_key_file = settings.ssh_user_ca_public_key_file
+      @client_ca_cert_file = settings.ssh_ca_cert_file
+      @client_cert_file = Proxy::RemoteExecution::Ssh.cert_file
```
lhellebr, line 66: Why do you treat these inconsistently? There is also `Proxy::RemoteExecution::Ssh.ca_public_key_file` but you don't use it. And for the others, there is no such method.

lib/smart_proxy_remote_execution_ssh/multiplexed_ssh_connection.rb
```diff
@@ -60,6 +60,10 @@ def initialize(options, logger:)
       @host_public_key = options.fetch(:host_public_key, nil)
       @verify_host = options.fetch(:verify_host, nil)
       @client_private_key_file = settings.ssh_identity_key_file
+      @client_ca_known_hosts_file = settings.ssh_ca_known_hosts_file
```
lhellebr, line 63: I'm confused by `ca` in name. This is a path to the known_hosts file where we expect to find a host-pubkey matching, right? What does it have to do with CA?

lib/smart_proxy_remote_execution_ssh/multiplexed_ssh_connection.rb
```diff
@@ -154,9 +158,14 @@ def establish_ssh_options
...
       ssh_options << "-o Port=#{@ssh_port}" if @ssh_port
       ssh_options << "-o IdentityFile=#{@client_private_key_file}" if @client_private_key_file
+      ssh_options << "-o CertificateFile=#{@client_cert_file}" if @client_cert_file
       ssh_options << "-o IdentitiesOnly=yes"
       ssh_options << "-o StrictHostKeyChecking=accept-new"
-      ssh_options << "-o UserKnownHostsFile=#{prepare_known_hosts}" if @host_public_key
+      if @host_public_key
+        ssh_options << "-o UserKnownHostsFile=#{prepare_known_hosts}"
+      elsif @client_ca_known_hosts_file
+        ssh_options << "-o UserKnownHostsFile=#{@client_ca_known_hosts_file}"
```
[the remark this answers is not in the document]
lhellebr: IMO yes. Otherwise, we get "if you trust this cert, connect and if you don't, connect anyway".

lib/smart_proxy_remote_execution_ssh/plugin.rb
```diff
@@ -12,6 +12,8 @@ class Plugin < Proxy::Plugin
 
     settings_file "remote_execution_ssh.yml"
     default_settings :ssh_identity_key_file   => '~/.ssh/id_rsa_foreman_proxy',
+                     # :ssh_ca_known_hosts_file => nil,
+                     # :ssh_user_ca_public_key_file  => nil,
```
lhellebr, line 16: Remove this?
Or uncomment, but then there's `ssh_ca_cert_file` missing.

lib/smart_proxy_remote_execution_ssh/multiplexed_ssh_connection.rb
adamlazik1: It relates to my question to the comment above: https://github.com/theforeman/smart_proxy_remote_execution_ssh/pull/126#discussion_r2199775536

cert_file is, for now,  a non-editable value whose name depends on the identity key file, so it is not in settings. Personally, I do not see any problem to making the cert file name an editable value, I just wanted to confirm first whether there are any objections against it.

lib/smart_proxy_remote_execution_ssh/multiplexed_ssh_connection.rb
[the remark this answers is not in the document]
adamlazik1: I too would say that row 4 could fail, but there is currently option to only have one host CA per smart proxy so it would be setting behavior for all hosts that use that proxy for REX. That is probably the only point that I see why it should fallback to plain SSH key instead of fail.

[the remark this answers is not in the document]
adamlazik1: Until we decide on one or the other, I will treat the expected behavior of row 4 to be failure. I updated the StrictHostKeyChecking param accordingly.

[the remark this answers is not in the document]
adamruzicka: So row 4 should indeed result in a failure, added two more rows to the table describing one other case that we previously missed.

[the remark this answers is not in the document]
adamlazik1: I can confirm that rows 7 and 8 also behave as intended, so the behavior of the whole feature should work as expected.

lhellebr (approved):

adamruzicka (commented):
one last nitpick, otherwise looking good

adamruzicka (approved):
Looks good

PR #131, opened 2026-03-05
Fixes #39132 - Do not rely on ACLs where not strictly necessary

pondrejk:
cases:
ssh user unprivileged + effective user unprivileged --> use setfacl as before
ssh user root  + effective user unprivileged --> use ssh user privileges to set via chmod
ssh user unprivileged + effective user root --> no need for extra privileges

### DIFF
--- a/lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
+++ b/lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
@@ -420,8 +420,13 @@ module Proxy::RemoteExecution::Ssh::Runners
     end
 
     def ensure_effective_user_access(*paths, mode: 'rx')
-      unless @user_method.is_a? NoopUserMethod
-        ensure_remote_command("setfacl -m u:#{@user_method.effective_user}:#{mode} #{paths.join(' ')}")
+      return if @user_method.is_a?(NoopUserMethod) || @user_method.effective_user == 'root'
+
+      paths_str = paths.join(' ')
+      if @user_method.ssh_user == 'root'
+        ensure_remote_command("chown #{@user_method.effective_user} #{paths_str} && chmod u=#{mode} #{paths_str}")
+      else
+        ensure_remote_command("setfacl -m u:#{@user_method.effective_user}:#{mode} #{paths_str}")
       end
     end
   end


ofedoren (commented):
Just a few cents: this fix explicitly checks if the user is `root`, but what about users that are sudoers, but are not `root`?

The steps in the issue suggest that it shouldn't fail for sudoers, but that's not that clear, thus asking.

lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
```diff
@@ -420,8 +420,13 @@ def check_expecting_disconnect
     end
 
     def ensure_effective_user_access(*paths, mode: 'rx')
-      unless @user_method.is_a? NoopUserMethod
-        ensure_remote_command("setfacl -m u:#{@user_method.effective_user}:#{mode} #{paths.join(' ')}")
+      return if @user_method.is_a?(NoopUserMethod) || @user_method.effective_user == 'root'
```
adamruzicka, line 423: In case where `@user_method.effective_user == 'root' && @user_method.ssh_user != 'root'`, does the ssh user have permissions to clean up all the working directories around L256 when the job is done?

pondrejk: hm, well using the scenario from the downstream tests it does, but I suppose that's because it is in wheel, so I guess it would depend on the ssh_user rights. But this was the case before this change as well.

adamruzicka: > that's because it is in wheel

Ohh, so the ownership transfers to the effective user, but the group ownership is left to the connection user's primary group, meaning the connection user retains write permissions through the group, cool. So not necessarily because of wheel, but because of a group

pondrejk:
prt seem happy at the moment in https://github.com/SatelliteQE/robottelo/pull/20950

adamruzicka:
> Just a few cents: this fix explicitly checks if the user is root, but what about users that are sudoers, but are not root?

Whether a user is a member of a specific group shouldn't really matter. This is somewhat configuration-dependant, but membership in sudoers/wheel only grants the user permissions to become another user, but until they change to be a different user, they're still an unprivileged user as any other.

adamruzicka (approved):

adamruzicka:
Thank you @pondrejk !

PR #132, opened 2026-05-18
Fix Encoding::CompatibilityError in password filtering

adamruzicka:
The SSH stream always returns ASCII-8BIT data, which is incompatible
with the UTF-8 regexp built from the password when it contains
wide characters. Match in binary encoding on both sides to avoid this.

Co-authored-by: Cursor <cursoragent@cursor.com>

### DIFF
--- a/lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
+++ b/lib/smart_proxy_remote_execution_ssh/runners/script_runner.rb
@@ -22,7 +22,8 @@ module Proxy::RemoteExecution::Ssh::Runners
     end
 
     def filter_password?(received_data)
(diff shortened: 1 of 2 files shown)


adamruzicka:
Tests should turn green after https://github.com/theforeman/smart_proxy_remote_execution_ssh/pull/133

pondrejk (approved):

pondrejk (approved):

ofedoren:
Could you please rebase?

adamruzicka:
rebased

ofedoren (approved):
Thanks, @adamruzicka and @pondrejk !

